PR #152, opened 2024-02-04
Admin UI improvements

Andrew-Dickinson:
- Adds mesh logo to the header
- Adds warning text about read-only nature of the db
- Add favicons to `meshweb` and the admin UI


Andrew-Dickinson:
Screenshot 2024-02-03 at 20 08 56
Screenshot 2024-02-03 at 20 09 14

WillNilges (approved):
Nice. We should also update the color scheme. That blue clashes something awful. I wonder how a yellow or orange would look.

PR #158, opened 2024-02-08
Add KML generator endpoint & example KML files

Andrew-Dickinson:
Adds a kml generator, to reproduce the behavior of the apps-script hosted one. I reverse engineered the apps script logic and checked against the output files, so this should produce KML files that are isomorphic to the originals. Once it's up we can send one to Olivier for verification and feedback

Screenshot 2024-02-08 at 00 25 48

### DIFF
--- a/src/meshapi/views/geography.py
+++ b/src/meshapi/views/geography.py
@@ -1,0 +1,171 @@
+from typing import Dict
+
+from django.db.models import F, Q
+from django.db.models.functions import Greatest
+from django.http import HttpResponse
+from fastkml import Data, ExtendedData, geometry, kml, styles
+from fastkml.enums import AltitudeMode
+from pygeoif import LineString, Point
+from rest_framework import permissions, status
+from rest_framework.decorators import api_view, permission_classes
+
+from meshapi.models import Building, Install, Link
+
+KML_CONTENT_TYPE = "application/vnd.google-earth.kml+xml; charset=utf-8"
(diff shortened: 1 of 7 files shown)


src/meshapi/views/geography.py
```diff
@@ -0,0 +1,171 @@
...
+    "Staten Island": "Staten Island",
+    None: "Other",
+}
+
+
+@api_view(["GET"])
+@permission_classes([permissions.AllowAny])
+def map_kml(request):
+    kml_root = kml.KML()
+    ns = "{http://www.opengis.net/kml/2.2}"
```
WillNilges, line 34: Just a thought, we should document all our external dependencies somewhere 🤔

Andrew-Dickinson: Sure, but this isn’t a dependency. This is just a namespace for the XML schema that KML uses. We don’t call this at runtime

WillNilges: oh, lol

WillNilges (approved):
Wow, this is sick. Should we put a frontend on meshforms?

Andrew-Dickinson:
If by frontend you mean "a link to the KML download" then sure haha

Andrew-Dickinson:
Whoops this was merged to the wrong place. Cherry picked this onto main via f1126795c1fbe375ae29fb403016eb3955f4fc09

Issue #93, opened 2023-12-24
Email based de-duplication carries data loss risks / requires operational changes


Andrew-Dickinson:
Currently there are many installs under `brian@nycmesh.net` in the spreadsheet data, many of them Grand St installs with real people's names who don't have email addresses. It also seems to pop up with some orgs (Riseboro, HPD, etc)

Currently if you try to sign one of those up new, we'd throw away the name and phone number because we'd think it's the "same person" because the email matches. 

This also poses a problem for name changes, etc. Since we will just drop the person's new information in the current implementation

Possible solutions:
- De-duplicate on (Email, Phone) pairs (since most of the grand st folks have telephones)
- Create a special case for `brian@nycmesh.net`
- Add a special mode for these kinds of signups, which an admin can use to create a "member" with no email
- Keep track of "requested member changes" in a queue and have the admin resolve them manually

Andrew-Dickinson:
Notes from call today:
- Deupe on phone, email pairs
- Set up meshforms to keep track of every submission
- Send conflict notifications to a slack channel

Andrew-Dickinson:
Notes from call with Willard today:
- Either email or phone number or both from join form
    - Dedupe with email if we get it
    - Dedupe with phone if not
- If a change is detected -> send slack message
- Meshforms needs to log every submission
- We need to do address autocomplete on the meshforms frontend

Andrew-Dickinson:
See also: #438

PR #160, opened 2024-02-09
Fix: Race condition for simultaneous NN assignment requests

Andrew-Dickinson:
A race condition existed when two NN assignment requests are received concurrently, where both requests would get the same NN. This was probably never gonna happen, but once I wrote a test case confirming it was possible I had to fix it :)

### DIFF
--- a/src/meshapi/tests/test_nn.py
+++ b/src/meshapi/tests/test_nn.py
@@ -1,11 +1,15 @@
 import json
+import threading
+import time
+from unittest import mock
 
 from django.conf import os
 from django.contrib.auth.models import User
-from django.test import Client, TestCase
+from django.test import Client, TestCase, TransactionTestCase
 
 from meshapi.models import Building, Install, Member
 
+from ..views import get_next_free_nn
 from .group_helpers import create_groups
 from .sample_data import sample_building, sample_install, sample_member
 
@@ -59,7 +63,7 @@ class TestNN(TestCase):
         self.assertEqual(
             expected_nn,
             resp_nn,
-            f"status code incorrect for test_nn_valid_install_number. Should be {code}, but got {response.status_code}",
+            f"nn incorrect for test_nn_valid_install_number. Should be {expected_nn}, but got {resp_nn}",
         )
 
         # Now test to make sure that we get 200 for dupes
@@ -81,7 +85,7 @@ class TestNN(TestCase):
         self.assertEqual(
             expected_nn,
             resp_nn,
-            f"status code incorrect for test_nn_valid_install_number. Should be {code}, but got {response.status_code}",
+            f"nn incorrect for test_nn_valid_install_number. Should be {expected_nn}, but got {resp_nn}",
         )
 
     def test_nn_invalid_password(self):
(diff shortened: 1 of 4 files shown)


src/meshapi/tests/test_nn.py
```diff
@@ -59,7 +63,7 @@ def test_nn_valid_install_number(self):
         self.assertEqual(
             expected_nn,
             resp_nn,
-            f"status code incorrect for test_nn_valid_install_number. Should be {code}, but got {response.status_code}",
+            f"nn incorrect for test_nn_valid_install_number. Should be {expected_nn}, but got {resp_nn}",
```
WillNilges, line 66: Oh, uh, good catch?

src/meshapi/views/forms.py
```diff
@@ -191,13 +190,44 @@ def join_form(request):
...
+    # This code may look useless but is extremely important. Here we use select_for_update()
+    # to acquire locks on every single row in the Install table. This is a bit hacky, but based
+    # on how our NN assignment logic works, we kind of need it. There's nothing else we can lock
+    # to prevent the same NN from being assigned to multiple installs when requested concurrently.
+    # We order by the table's primary key to prevent deadlocks due to out-of-order locking, and we
+    # call len() to make sure the queryset is traversed (which is what actually acquires the lock)
+    # This must be called in the context of a transaction.atomic block, which also automatically
+    # handles the release of the lock for us
+    lock_qs = Install.objects.select_for_update().all().order_by("install_number")
+    len(lock_qs)
```
WillNilges, line 220: Very thorough explanation. I was gonna posit adding something like this to the join form, but that might not be necessary since the likelihood of collisions is so low. (Mostly thinking about what would happen if two members in the same building submitted the join form at the same time.) Thoughts?

WillNilges (approved):
Nice

src/meshapi/tests/test_nn.py
Andrew-Dickinson: Haha yeah I copied this and it wasn't right

src/meshapi/views/forms.py
Andrew-Dickinson: This isn't necessary on the join form, since the database handles the atomicity for us. Since the database automatically assigns the `install_number` and that is the table's primary key, it knows not to let two rows have the same value.

However, it is a good idea to confirm this with a test, so I'll add one

Andrew-Dickinson: Oh wait I see now, you're talking about simultaneous same-building signups. Hmmm

Andrew-Dickinson: Okay I add locks to the join form also

src/meshapi/views/forms.py
```diff
@@ -214,8 +262,14 @@ def network_number_assignment(request):
         print(f"NN Request failed. Could not decode request: {e}")
         return Response({"detail": "Got incomplete request"}, status=status.HTTP_400_BAD_REQUEST)
 
+    # This call is a bit out of order, but we need to do this here so that we acquire the lock
+    # before we call Install.objects.select_for_update().get() below in order to prevent deadlocks
+    free_nn = get_next_free_nn()
+
     try:
-        nn_install = Install.objects.get(install_number=r.install_number)
+        # We don't need select_for_update here because our call above acquires a lock on every row,
```
Andrew-Dickinson, line 270: Whoops this comment is wrong now

src/meshapi/views/forms.py
```diff
@@ -191,13 +215,37 @@ def join_form(request):
     )
 
 
+def get_next_free_nn() -> int:
+    # We don't need to lock all the DB objects, just the ones that impact the free NN calculations
```
Andrew-Dickinson, line 219: Could we get away with just the first one, join form style?

src/meshapi/views/forms.py
```diff
@@ -35,8 +37,25 @@ class JoinFormRequest:
     ncl: bool
 
 
+def acquire_lock_for_queryset(queryset: QuerySet, object_count=None):
+    # This code may look useless but is extremely important. Here we use select_for_update()
+    # to acquire locks on every single row in the given table. This is a bit hacky, but based
```
Andrew-Dickinson, line 42: Fix me: this isn’t alway every row anymore

Andrew-Dickinson:
Idk locking a random arbitrary row works for this but just feels bad. There’s got to be a better way

src/meshapi/views/forms.py
WillNilges: What are the implications of that?

src/meshapi/views/forms.py
```diff
@@ -88,6 +108,11 @@ def join_form(request):
             {"detail": "Your address could not be validated."}, status=status.HTTP_500_INTERNAL_SERVER_ERROR
         )
 
+    # No need to lock all DB objects since we don't modify these tables,
```
WillNilges, line 111: We don't? Just to be clear, do you mean to say that all we do is append? What if we get a response from the same Member with different info? IIRC, we update their information, don't we?

src/meshapi/views/forms.py
WillNilges: > Just the ones that impact the free NN calculations
Do you have an example?

> Just the first one
Not totally clear on what locking the first one does.

src/meshapi/views/forms.py
```diff
@@ -244,12 +262,14 @@ def network_number_assignment(request):
         print(f"NN Request failed. Could not decode request: {e}")
         return Response({"detail": "Got incomplete request"}, status=status.HTTP_400_BAD_REQUEST)
 
-    acquire_lock_for_all_installs()
+    # This call is a bit out of order, but we need to do this here so that we acquire the lock
+    # before we call Install.objects.select_for_update().get() below in order to prevent deadlocks
+    free_nn = get_next_free_nn()
```
WillNilges, line 267: This does so to lock as soon as we know the request is valid? Makes sense, because I don't think that `get_next_free_nn` has any side effects.

WillNilges (commented):
I suppose I'd like to have bit of a conversation about the async stuff before I approve it just so I understand the architecture better, and to make sure that we don't do something that would shoot us in the foot later.

Andrew-Dickinson:
Yeah agreed. I think there might be room here for a better strategy. Now
that we have the tests to reproduce the race condition we can easily
evaluate different locking options

Andrew Dickinson
Computer Science BSE (December 2020) - University of Michigan

src/meshapi/views/forms.py
Andrew-Dickinson: Currently, we do not update their information, to prevent griefing. We have an open issue about this: #93

Andrew-Dickinson:
Okay this is overhauled to move away from hacky-row based locks and use a postgres feature that is _specifically_ intended for this purpose. I feel a lot better about the robustness of this, and the code is a lot cleaner too :)

WillNilges (approved):

Issue #73, opened 2023-09-24
Add `notes` TextField and a `DIY` boolean to install model


WillNilges:

Andrew-Dickinson:
The notes field is present. Are we still interested in the `DIY` bool?

Andrew-Dickinson:
I don't believe we don't have this information currently, so maybe we'd have to assume that all installs are not DIY?

Andrew-Dickinson:
Actually I guess looking for the text "DIY" in the notes field is a pretty good signal

Andrew-Dickinson:
#161

PR #161, opened 2024-02-09
Add DIY field

Andrew-Dickinson:
Adds a DIY field to the install model. This is populated based on the text "diy" appearing anywhere in the imported notes, and is set to `null` for installs that never got installed


Andrew-Dickinson:
Closes #73

WillNilges (approved):
Love it. Is this something we want to expose to the Member on the Join Form, or should we leave it up to Olivier on the Join Form?

One Nit: Please make sure that the checkbox shows up in a reasonable place on the Admin Panel, and if it does not, at least make an issue for it!!!

Andrew-Dickinson:
Oh good call. Because we're using a custom admin form, it's actually not present by default. I will add it

Andrew-Dickinson:
My thought was yeah that this ends up being admin-maintained metadata. DIYs are rare so that should be no problem

PR #171, opened 2024-02-10
Add build + publish workflow

WillNilges:
Builds an image, publishes it to Docker Hub, then deploys it on the grandbox.

### DIFF
--- a/.github/workflows/publish-and-deploy.yaml
+++ b/.github/workflows/publish-and-deploy.yaml
@@ -46,4 +46,4 @@ jobs:
           #config: ${{ secrets.CONFIG }} # ssh_config; optional
           if_key_exists: fail # replace / ignore / fail; optional (defaults to fail)
       - name: Pull new Docker image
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-and-deploy.yaml
```diff
@@ -0,0 +1,51 @@
...
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@9ec57ed1fcdbf14dcef7dfbe97b2010124a938b7
+        with:
+          images: willnilges/meshdb
+      
+      - name: Build and push Docker image
+        uses: docker/build-push-action@3b5e8027fcad23fda98b2e3ac259d8d67585f671
```
Andrew-Dickinson, line 29: Does this publish the image publically? Are there any risks of credentials being "baked into" the image? I think we're using runtime environment variables for everything but just wanted to double check.

Support bot bakes in creds at build time I'm pretty sure (so we have to keep the builds private)

Andrew-Dickinson (approved):

PR #176, opened 2024-02-10
Add back NN Assigned status - needed for proper map rendering

Andrew-Dickinson:
Turns out we need to keep track of when an install has its NN reassigned to someone else, so that we don't double-render it on the map. This PR adds it back so that we can use it in a follow up PR to fix the map


src/meshapi/views/forms.py
```diff
@@ -209,6 +209,34 @@ def get_next_free_nn() -> int:
...
+            nn_donor_install.install_status != Install.InstallStatus.REQUEST_RECEIVED
+            or nn_donor_install.network_number is not None
+        ):
+            raise ValueError(
+                f"Invalid NN: {free_nn} has an install associated that "
+                f"looks active (#{nn_donor_install.install_number})"
+            )
+
+        nn_donor_install.install_status = Install.InstallStatus.NN_ASSIGNED
+        nn_donor_install.save()
```
WillNilges, line 238: Why not simply leave the NN field blank and use another status like `CLOSED` or something? This information doesn't seem that useful, and this seems completely irrelevant for installs going forward.

src/meshapi/views/forms.py
```diff
@@ -195,7 +195,7 @@ def join_form(request):
     )
 
 
-def get_next_free_nn() -> int:
+def get_next_free_nn_from_old_installs() -> int:
     defined_nns = set(
         Install.objects.exclude(install_status=Install.InstallStatus.REQUEST_RECEIVED, network_number__isnull=True)
```
WillNilges, line 200: Just checking: Should we be excluding `REQUEST_RECEIVED`? Do we know for sure at this point in the code that no `REQUEST_RECEIVED` Installs have NNs?

src/meshapi/views/forms.py
```diff
@@ -195,7 +195,7 @@ def join_form(request):
     )
 
 
-def get_next_free_nn() -> int:
+def get_next_free_nn_from_old_installs() -> int:
```
WillNilges, line 198: Why the name change? Maybe it would be better to have a separate function?

WillNilges (commented):
Maybe I misunderstand, but this seems like a change purely to accommodate the import script? I am not clear on why we need to have this? Couldn't we do the same thing by simply stripping out the NN from installs that are from the spreadsheet who have had their NN reassigned?

Speaking of reassigned, could we at least refactor from `NN_ASSIGNED` to `NN_REASSIGNED` or something? The name is really confusing.

src/meshapi/views/forms.py
Andrew-Dickinson: `get_available_nn`

WillNilges (approved):

WillNilges:
Please change the function name and change the status to `REASSIGNED_AS_NN`

WillNilges:
We had a phonecall

Issue #163, opened 2024-02-09
Validate GIS data


Andrew-Dickinson:
We need to validate that our imported data looks right when viewed as a map. I think right away we probably want to change how lat/lon import works, since right now we're throwing away manual spreadsheet tweaks. We should preserve these

Andrew-Dickinson:
Took a first stab at this via #188, but we need to wait for a deploy to ensure things actually look right

Andrew-Dickinson:
More good progress in #191 things look pretty good in local dev, there are only some small details that look off due to known data model limitations (we consolidate all installs at the same building into a single point, e.g. Vernon APs). I made a [slack post](https://nycmesh.slack.com/archives/C06GW3Q92TY/p1707890569913499?thread_ts=1707029793.632389&cid=C06GW3Q92TY) about this situtation, since really this needs to be addressed as part of #166, this isn't a GIS issue.

I think once we merge #191, I will do one last check that everything else except the APs look good and close this out

PR #177, opened 2024-02-10
Map endpoint audit fixes

### LINKED ISSUE #164: Audit /mapdata endpoints against the logic of the google apps script
Now that I know where to look, validate that the logic is the same for our map endpoint clones

Andrew-Dickinson:
Audited the map endpoint code and compared against the node-db version: https://github.com/nycmeshnet/node-db/blob/master/src/spreadsheet.js#L119

Found a few discrepencies:
 - Nodes from re-assigned NNs are not being handled correctly
 - Abandoned sectors should not be included in the output JSON
 
This PR fixes these issues and closes #164

### DIFF
--- a/src/meshapi/serializers/map.py
+++ b/src/meshapi/serializers/map.py
@@ -88,6 +88,8 @@ class MapDataInstallSerializer(serializers.ModelSerializer):
             return "Powered Off"
         elif install.install_status == Install.InstallStatus.CLOSED:
             return "Abandoned"
+        elif install.install_status == Install.InstallStatus.NN_REASSIGNED:
+            return "NN Assigned"
 
         return install.install_status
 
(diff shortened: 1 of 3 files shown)


WillNilges:
Why are we using NodeDB as a source of truth? Why not verify against the AppScript directly?

> Nodes from re-assigned NNs are not being handled correctly

Can you elaborate?

Andrew-Dickinson:
Today, the map is composed of a JSON payload that depends on both NodeDB and appscript code. NodeDB calls the appscript endpoint to get a list of all installs, links, and sectors and then filters them via the logic I linked above. The audit included both codebases, and we are now roughly compliant with their logic

However, there are definitely some bugs in our system, the display doesn't look right, but that will be tracked via #163 since it is likely broader than just the `/mapdata` endpoints (e.g. it includes import script related fixes also)

WillNilges (approved):

WillNilges:
We had a phonecall

PR #188, opened 2024-02-13
Address handling fixes in import script

Andrew-Dickinson:
A few fixes that help with how we handle addresses:

- Fix common typos like "steet" and "avue" that confuse Pelias
- Use spreadsheet coordinates instead of our own computed coordinates (to include admin tweaks from the spreadsheet in our imported data)
- Add warning to `dropped_edits.csv` and building notes field when the spreadsheet lat/lon is more than 100 meters from our computed value based on the address
- Fix: "Bowery" confuses Pelias


WillNilges (approved):
Talked about it on a phonecall:
- Bowery Hack
- Small refactor
- Has unit tests(!)

Issue #192, opened 2024-02-15
Bug: APs are squashed on top of the building they are located at


Andrew-Dickinson:
Currently the APs at Vernon all have the same NN (5916), BIN (3042881), and address (303 Vernon Ave, Brooklyn) as Vernon itself, but there are install entries (e.g. 11875, 11876, 11877, etc.)  with different lat/lon values for each AP. Right now our import process consolidates them by BIN, which results in them all being squished on top of Vernon. The same thing is happening at 1350, the installs 3736, 3737, 3738, and 3739 have been “spread out” across the rooftop, but since they all have the same address and BIN we consolidate them into a single building and loose that information (in this case this is probably acceptable, less so in the AP case)

Andrew-Dickinson:
1350 Actual
<img width="892" alt="Screenshot 2024-02-14 at 23 50 33" src="https://github.com/WillNilges/meshdb/assets/3138937/3e998a09-7052-4745-8e0d-6499e325a44a">

1350 Expected
<img width="892" alt="Screenshot 2024-02-14 at 23 51 18" src="https://github.com/WillNilges/meshdb/assets/3138937/e11f7472-31e0-4b7f-8280-b71b8ab1e46e">

Andrew-Dickinson:
Vernon Actual
<img width="892" alt="Screenshot 2024-02-14 at 23 50 15" src="https://github.com/WillNilges/meshdb/assets/3138937/73ec5087-a394-4cb4-a2c4-607e255cb5c8">

Vernon Expected
<img width="892" alt="Screenshot 2024-02-14 at 23 51 04" src="https://github.com/WillNilges/meshdb/assets/3138937/34dce1a7-d097-4d34-b0f6-0815dab031d8">

Andrew-Dickinson:
Resolved via #228

PR #193, opened 2024-02-15
Adjust link serialization logic to match current map display

Andrew-Dickinson:
Adjust the serialization logic so that we line up nearly perfectly with the current map display (and a few small refactors). There are some small remaining discrepancies some are due to bad spreadsheet data (that we validate and correct at import time, resulting in nodes changing slightly), and AP/multi-install display (captured in #192) but 95% of the map looks identical.

Here's an example. There's an abandoned node (NN558) in the spreadsheet data that shows up on the map incorrectly, we've filtered it out

Map based on our data:
Screenshot 2024-02-15 at 00 12 42

Live map
Screenshot 2024-02-15 at 00 13 52

### DIFF
--- a/src/meshapi/serializers/map.py
+++ b/src/meshapi/serializers/map.py
@@ -5,6 +5,12 @@ from rest_framework import serializers
 
 from meshapi.models import Install, Link, Sector
 
+EXCLUDED_INSTALL_STATUSES = {
+    Install.InstallStatus.CLOSED,
+    Install.InstallStatus.NN_REASSIGNED,
+}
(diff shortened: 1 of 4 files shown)


Andrew-Dickinson:
This closes #163 since the map now looks good

WillNilges (approved):
LGTM, a couple of questions about how the current map works:
- Are there inconsistencies with how abandoned nodes are treated?
- Why were/are abandoned nodes shown? If we are changing that (sounds like we are), what led you to this decision?

PR #196, opened 2024-02-16
Fix: Member search responds with HTTP 500

quoncc:
Typo in admin.py MemberAdmin class was causing searching by member to error out


Andrew-Dickinson (approved):
Quincy you're alive :D

PR #199, opened 2024-02-16
Add link and sector endpoints

Andrew-Dickinson:
Add endpoints, serializers, etc to allow API access to create, read, update, and delete `Link` and `Sector` objects. Includes test coverage, including an absolute unit of a test result that shows the huge payloads this recursive behavior can create

PR #200, opened 2024-02-16
Add the ability for clients to limit recursion depth to reduce response sizes

Andrew-Dickinson:
The chonker of a response object introduced in #199 made me think that some clients are going to want the ability to limit the depth of the recursion in the object they receive, this PR adds a `?max_recursion_depth` parameter available on any of the endpoints that are backed by `RecursiveSerializer` which can be optionally used to limit the depth if the caller is only interested in shallower data fields


Andrew-Dickinson:
DON'T MERGE ME WITHOUT MERGING #199 AND CHANGING THE DESTINATION BRANCH YOU DUMMY

PR #204, opened 2024-02-18
Remove unused /user endpoints to reduce confusion

Andrew-Dickinson:
Remove user endpoints, these are not needed and just serve to confuse our API consumers, given the close naming but totally separate purpose to the `Member` table


WillNilges (approved):
:shipit:

PR #208, opened 2024-02-19
Remove recursive serialization

Andrew-Dickinson:
Removes recursive serialization, and modifies tests to expect `pk` integer fields for foreign keys instead of inline objects


Andrew-Dickinson:
Closes #207

Issue #202, opened 2024-02-18
Buildings without a street address aren't editable


Andrew-Dickinson:
Street address is a strongly recommended, but not a required field (because of import shenanigans)

We have buildings (inactive, but in the DB nonetheless) that do not have a street address, and they are not editable

Steps to reproduce:
1. Create a new building without a street address
2. Try to edit the building from the list of buildings (there is no link to click)

![Screenshot 2024-02-18 at 12 57 59](https://github.com/WillNilges/meshdb/assets/3138937/8791d005-8d1c-41e7-9b75-08c1d8489a0e)

Issue #112, opened 2024-01-23
Improve interface for updating relations on Admin Panel.


WillNilges:
![Untitled](https://github.com/andybaumgar/meshdb/assets/42927786/a8cec1b5-1125-440a-80fa-b0d1e04fa0b1)

Andrew-Dickinson:
This is most of the way there, but I think we still need some work on "reverse" relations. For example when adding an install to a building it's impossible to select a pre-existing install (instead you can only create a new one)

Andrew-Dickinson:
Also, I don't think we want to encourage admins to add installs in that way, can we disable that option altogether?

quoncc:
Not 100% sure what the issue is? This UI looks fine now

<img width="451" alt="image" src="https://github.com/WillNilges/meshdb/assets/76627323/ef0bc250-8bdc-427d-93f4-49bd7ef1a88b">

Andrew-Dickinson:
Right, WIllard has done most of this, so the original issue text is a bit out of date. The remaining work is described by my comment above: 

> I think we still need some work on "reverse" relations. For example when adding an install to a building it's impossible to select a pre-existing install (instead you can only create a new one)

PR #209, opened 2024-02-19
Update list views to avoid blank entries, fix TabularInline Install view, and other bugfixes

WillNilges:
Right now, we've got some problems with buildings and such not having street names, which are used as the primary title in the Admin Panel, Causing them to be unselectable. See #202.

This'll make the List View always use "__str__" in the models, which should guarantee that _something_ shows up for the row.

Side effect of this: Sometimes the information in the following columns can be redundant.

Example: 
image

Changes:
- Buildings, instead of always using their street name as primary view, will now use Building Name, falling back on Street Address, BIN, and finally MeshDB Building ID.
- Links, instead of using MeshDB Link IDs, will use `NN to NN`, falling back on MeshDB Link IDs.
- Sectors, instead of using MeshDB Sector IDs, will use SSID, falling back on MeshDB Sector IDs.

### DIFF
--- a/src/meshapi/admin.py
+++ b/src/meshapi/admin.py
@@ -9,12 +9,13 @@ admin.site.index_title = "Welcome to MeshDB Admin Portal"
 
 # Register your models here.
 
+
 class InstallInline(admin.TabularInline):
     model = Install
     extra = 0  # Number of empty forms to display
     show_change_link = True
-    fields = ["install_status", "network_number", "member", "unit"] 
-    readonly_fields = fields 
+    fields = ["install_status", "network_number", "member", "unit"]
+    readonly_fields = fields
     can_delete = False
 
     def has_add_permission(self, request, obj):
--- a/src/meshapi/admin.py
+++ b/src/meshapi/admin.py
@@ -163,6 +163,7 @@ class MemberAdmin(admin.ModelAdmin):
     ]
     inlines = [InstallInline]
     list_display = [
+        "__str__",
         "name",
         "primary_email_address",
         "phone_number",
@@ -268,7 +269,7 @@ class LinkAdmin(admin.ModelAdmin):
         "from_building__icontains",
         "to_building__icontains",
     ]
-    list_display = ["id", "from_building", "to_building"]
+    list_display = ["__str__", "from_building", "to_building"]
     list_filter = ["status", "type"]
 
 
--- a/src/meshapi/admin.py
+++ b/src/meshapi/admin.py
@@ -9,10 +9,16 @@ admin.site.index_title = "Welcome to MeshDB Admin Portal"
 
 # Register your models here.
 
-
 class InstallInline(admin.TabularInline):
     model = Install
-    extra = 1  # Number of empty forms to display
+    extra = 0  # Number of empty forms to display
+    show_change_link = True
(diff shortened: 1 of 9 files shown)


WillNilges:
Blocked waiting on more spreadsheet data so I can load the Members and install tables again 😿

WillNilges:
This now also partially addresses #112 by updating the Install Inline in the Building Admin

image

WillNilges:
OK, here's how everything looks now:

Installs under building:

image

Install List View:
image

Link List View:
image


Sector List View:
image

Installs under Member:

image

src/meshapi/tests/test_admin_list_view.py
```diff
@@ -0,0 +1,34 @@
+from django.test import Client, TestCase
+from django.contrib.auth.models import User
+
+
+# Sanity check to make sure that the list views in the admin panel still work
+# These will often break when you update something in the model and forget to
+# update the admin panel
+class TestAdminListView(TestCase):
```
Andrew-Dickinson, line 8: Can we please test search views as well? Currently list views work but searches don't

src/meshapi/models.py
```diff
@@ -193,3 +198,8 @@ class SectorStatus(models.TextChoices):
     ssid = models.TextField(default=None, blank=True, null=True)
 
     notes = models.TextField(default=None, blank=True, null=True)
+
+    def __str__(self):
+        if self.ssid:
+            return self.ssid
+        return f"MeshDB Sector ID {self.id}"
```
Andrew-Dickinson, line 205: Not sure if I agree with this logic. The omni "sectors" all have the same `Omni-wds` SSID so in that case it's not useful for telling them apart.

What about using `building.primary_nn` in addition to the SSID? I think there's good milage in the `name` and `device_name` fields also (which we have filled in for every sector)

src/meshapi/admin.py
```diff
@@ -148,7 +199,9 @@ class MemberAdmin(admin.ModelAdmin):
     search_fields = [
         # Search by name
         "name__icontains",
-        "email_address__icontains",
+        "primary_email_address__icontains",
+        "stripe_email_address__icontains",
+        "secondary_email_addresses__icontains",
```
Andrew-Dickinson, line 204: Ooops

src/meshapi/models.py
Andrew-Dickinson: This might be about to get totally changed in the schema revamp anyways so maybe it's fine for now

Andrew-Dickinson (approved):

src/meshapi/admin.py
WillNilges: I hope you're pleased with yourself >:(

WillNilges: And I can't use the `all_email_addresses` property to filter because the searches happen at the DB level.

https://stackoverflow.com/a/1205389

src/meshapi/models.py
WillNilges: There's not really a "clean" way to do this so that the list view looks pleasing and it's also totally unambiguous. I think SSID will get us closest to that. Other fields like `name` and `device name` seem to be even less unique.

Issue #71, opened 2023-09-21
Create documentation for how to use meshdb


WillNilges:
probably important to ease the learning curve as much as possible for people. This will be a totally new web UI for them, hopefully it's as simple and intuitive as possible, but some basic stuff like "how do I create a node number?" or "how do I do XYZ moderately complex thing through the admin UI?" or "what's the difference between an admin and an install leader?" or "how do I connect this to other services?" might be pertinent.

LydonThorpe:
Where do you want this documentation to live? Wiki?

Andrew-Dickinson:
We might end up hosting the docs for this on the site itself, just because that lets us do things like dynamically generate docs based on code content (javadoc style), it also lets you require doc changes in PRs that change the code, etc. However we can always link this content from the wiki, and certainly we should have some basic information over there

WillNilges:
TODO: Add help text on each of our fields.

WillNilges:
Looks like the API docs are live: https://db.grandsvc.mesh.nycmesh.net/api-docs/swagger/#/
and no longer loading infinitely.

Need to remove nonsense [api](https://db.grandsvc.mesh.nycmesh.net/api-docs/swagger/#/api) tab and add [Legacy Query Form](https://db.grandsvc.mesh.nycmesh.net/api-docs/swagger/#/Legacy%20Query%20Form)

Andrew-Dickinson:
Added a bunch of "helpdesk" style articles to the wiki: https://wiki.mesh.nycmesh.net/books/6-services-software/chapter/meshdb

I think this issue is good to resolve. We can supplement with additional information as we encounter gaps

PR #210, opened 2024-02-19
Implement API docs using drf-spectacular

Andrew-Dickinson:
I need to go to sleep. Here's a some screenshots:

Screenshot 2024-02-19 at 01 35 39
Screenshot 2024-02-19 at 01 36 01


Andrew-Dickinson:
Significant progress towards #71

src/meshapi/docs.py
```diff
@@ -0,0 +1,96 @@
...
+from drf_spectacular.types import OpenApiTypes
+from drf_spectacular.utils import OpenApiParameter, extend_schema
+from drf_spectacular.views import SpectacularSwaggerView
+
+
+class SpectacularSwaggerInjectVarsView(SpectacularSwaggerView):
+    @extend_schema(exclude=True)
+    def get(self, request, *args, **kwargs):
+        spectacular_settings.DESCRIPTION = (
+            "Programatic access to mesh core data, detailing our installs, members, etc. "
```
WillNilges, line 17: Programmatic*

src/meshapi/models.py
```diff
@@ -165,31 +327,58 @@ class SectorStatus(models.TextChoices):
...
+        on_delete=models.PROTECT,
+        related_name="sectors",
+        help_text="The building this sector is installed on",
+    )
+    name = models.TextField(
+        help_text="The name of the hub sector is installed on",
+    )
 
     radius = models.FloatField(
+        help_text="The radius to display this sector on the map (scale is km?)",
```
WillNilges, line 341: What do you mean, "scale is km?" Do we not know what unit this radius is in? We should figure that out (out of scope of this PR, sure, but might wanna make an issue for it).

src/meshapi/models.py
```diff
@@ -165,31 +327,58 @@ class SectorStatus(models.TextChoices):
...
+        null=True,
+        help_text="The SSID being broadcast by this device",
+    )
 
-    notes = models.TextField(default=None, blank=True, null=True)
+    notes = models.TextField(
+        default=None,
+        blank=True,
+        null=True,
+        help_text="A free-form text description of this Sector, to track any additional information.",
```
WillNilges, line 383: Might wanna add the disclaimer here.

```
However this structure can be changed by admins at any time and should not be relied on
```

src/meshapi/urls.py
```diff
@@ -24,10 +23,8 @@
     path("query/buildings/", views.QueryBuilding.as_view(), name="meshapi-v1-query-building"),
     path("query/members/", views.QueryMember.as_view(), name="meshapi-v1-query-member"),
     path("query/installs/", views.QueryInstall.as_view(), name="meshapi-v1-query-install"),
-    path("mapdata/installs/", views.MapDataInstallList.as_view(), name="meshapi-v1-map-data-installs"),
+    path("mapdata/nodes/", views.MapDataNodeList.as_view(), name="meshapi-v1-map-data-installs"),
```
WillNilges, line 26: Not super jazzed about canonizing a "nodes" route. I am assuming this is purely for backwards compatibility with the map, and any new map wouldn't do this.

I would argue for a "DISCLAIMER THIS IS FUCKING RADIOACTIVE DON'T FUCKING USE THIS" docstring somewhere.

src/meshapi/urls.py
```diff
@@ -24,10 +23,8 @@
...
     path("mapdata/links/", views.MapDataLinkList.as_view(), name="meshapi-v1-map-data-links"),
     path("mapdata/sectors/", views.MapDataSectorList.as_view(), name="meshapi-v1-map-data-sectors"),
-    path("geography/whole-mesh.kml", views.map_kml, name="meshapi-v1-geography-whole-mesh-kml"),
+    path("geography/whole-mesh.kml", views.WholeMeshKML.as_view(), name="meshapi-v1-geography-whole-mesh-kml"),
```
WillNilges, line 29: Interesting... I'm lazy. Can you give me a screenshot of this?

src/meshapi/views/forms.py
```diff
@@ -37,6 +39,42 @@ class JoinFormRequest:
...
+                ),
+                description="This install's building did not already have an install. One has been created and assigned.",
+            ),
+            "400": OpenApiResponse(
+                form_err_response_schema, description="Invalid request body JSON or missing required fields"
+            ),
+            "500": OpenApiResponse(form_err_response_schema, description="Unexpected internal error"),
+        },
+    ),
+)
```
WillNilges, line 77: Is this purely for documentation purposes?

src/meshapi/views/forms.py
```diff
@@ -37,6 +39,42 @@ class JoinFormRequest:
...
+                    "JoinFormSuccessResponse",
+                    fields={
+                        "detail": serializers.CharField(),
+                        "building_id": serializers.IntegerField(),
+                        "member_id": serializers.IntegerField(),
+                        "install_number": serializers.IntegerField(),
+                        "member_exists": serializers.BooleanField(),
+                    },
+                ),
+                description="This install's building did not already have an install. One has been created and assigned.",
```
WillNilges, line 69: We should clearly define the behavior of the join form here, and I don't think this is 100% accurate. It will create an Install basically as long as the Member has not created an install at the same building in the same unit before.

WillNilges:
```suggestion
                description="This install's building did not already have an install for this specific Member in this unit. The Member is deduplicated by Email address. The One has been created and assigned.",
```

src/meshapi/views/forms.py
```diff
@@ -255,6 +293,48 @@ class NetworkNumberAssignmentRequest:
...
+        "network_number": serializers.IntegerField(),
+        "created": serializers.BooleanField(),
+    },
+)
+
+
+@extend_schema_view(
+    post=extend_schema(
+        tags=["User Forms"],
+        summary="Assign a network number to a given Install object. Used by the NN Assignment form",
```
WillNilges, line 318: Worthwhile to link to source code and/or meshforms here?

src/meshapi/views/forms.py
```diff
@@ -255,6 +293,48 @@ class NetworkNumberAssignmentRequest:
...
+        request=NetworkNumberAssignmentRequestSerializer,
+        responses={
+            "200": OpenApiResponse(
+                nn_form_success_schema,
+                description="This install already has an NN, no action has been perfomed in the DB",
+            ),
+            "201": OpenApiResponse(
+                nn_form_success_schema,
+                description="This install's building did not already have an install. One has been created and assigned.",
```
WillNilges, line 327: Huh? I think this might be a mis-copypasta. The NN assign form cannot create installs.

src/meshapi/views/geography.py
WillNilges, line 1: Assuming these changes are to make `as_view()` work?

src/meshapi/views/lookups.py
WillNilges, line 1: I am pretty sure that the Query Form only works for a subset of these fields.

src/meshapi/views/model_api.py
```diff
@@ -14,58 +16,125 @@
 )
 
 
-# Home view
+@extend_schema_view(
+    get=extend_schema(
+        tags=["API Status"],
+        # responses=OpenApiTypes.STR,
```
WillNilges, line 22:
```suggestion
```

src/meshdb/settings.py
```diff
@@ -213,3 +215,38 @@
...
+    "TITLE": "MeshDB Data API",
+    "VERSION": "1.0.0",
+    "SERVE_INCLUDE_SCHEMA": False,
+    "TAGS": [
+        {"name": "API Status", "description": "Meta-endpoint to indicate API status"},
+        {"name": "Members", "description": "Members of the mesh and their contact details"},
+        {"name": "Buildings", "description": "Buildings on the mesh and their location data"},
+        {
+            "name": "Installs",
+            "description": "Installs, one corresponding to each household on the mesh or wishing to join",
```
WillNilges, line 229:
```suggestion
            "description": "Installs, one corresponding to each household that is either already on the mesh, or wishes to join the mesh",
```

src/meshdb/settings.py
```diff
@@ -213,3 +215,38 @@
...
+        },
+        {"name": "Links", "description": "Links between buildings"},
+        {"name": "Sectors", "description": 'Special antennas with broad coverage of a radial "slice" of land area'},
```
WillNilges, line 232: Perhaps we could link to some documentation here? I don't think the concept of sectors is intuitive for non-meshers or those not familiar with RF.

src/meshdb/settings.py
```diff
@@ -213,3 +215,38 @@
...
+        {"name": "Geographic & KML Data", "description": "Endpoints for geographic and KML data export"},
+        {
+            "name": "Website Map Data",
+            "description": "Endpoints used to power the nycmesh.net website map. "
+            "Uses a legacy data format, not recommended for new applications",
```
WillNilges, line 237: Can we emphasize "NOT RECOMMENDED FOR NEW APPLICATIONS" somehow?

src/meshdb/settings.py
```diff
@@ -213,3 +215,38 @@
...
+        },
+        {
+            "name": "Legacy Query Form",
+            "description": "Endpoints used to power the legacy docs query form. "
+            "Uses a legacy data format, not recommended for new applications",
```
WillNilges, line 242: Ditto

src/meshdb/templates/rest_framework/login_base.html
```diff
@@ -0,0 +1,12 @@
+{% extends "rest_framework/login_base.html" %}
+
+{% block branding %}
+   <h3 style="margin: 0 0 5px;">MeshDB Data API Login</h3>
+    <p  style="margin: 0 0 20px; font-size: 12px">
+    <i>
+        Admin UI credentials work here, but non "staff" users who don't have admin
```
WillNilges, line 7: Wait, do we need to give people an account so they can access our docs? I'd rather they be totally open. If that's a security concern, we should address the security concern so our docs can be open. Make it more accessible that way.

src/meshdb/templates/drf_spectacular/swagger_ui.html
```diff
@@ -0,0 +1,41 @@
...
+      {#var lockedIcon = document.querySelector(".svg-assets defs symbol#locked");#}
+      {#var unlockedIcon = document.querySelector(".svg-assets defs symbol#unlocked");#}
+      {#lockedIcon.id = "unlocked";#}
+      {#unlockedIcon.id = "locked";#}
+
+      // Borrowed from here
+      // https://github.com/tfranzel/drf-spectacular/issues/1099#issuecomment-1920345459
+      // The sessionID variable gets magically inserted via the template settings object in docs.py
+      ui.preauthorizeApiKey("Session ID", sessionID);
+  };
```
WillNilges, line 39: Can you condense this into a FIXME for now?

src/meshdb/urls.py
```diff
@@ -16,9 +16,22 @@
...
-    path("", include("meshweb.urls")),
+    path("api-docs/openapi3.json", SpectacularAPIView.as_view(), name="api-docs-schema"),
+    path(
+        "api-docs/swagger/",
+        SpectacularSwaggerInjectVarsView.as_view(
+            url_name="api-docs-schema"  # , template_name="drf_spectacular/swagger_ui.html"
+        ),
+        name="api-docs-swagger",
+    ),
+    path("api-docs/redoc/", SpectacularRedocView.as_view(url_name="api-docs-schema"), name="api-docs-redoc"),
```
WillNilges, line 36: Can we change `api-docs` to `docs`?

src/meshweb/templates/meshweb/index.html
```diff
@@ -95,11 +95,18 @@ <h1>Welcome to MeshDB!</h1>
...
       <a href="https://forms.grandsvc.mesh.nycmesh.net/join/">Join Our Community Network!</a>
-      <a href="/admin">Looking for the admin panel?</a>
+      <a href="/admin">Admin Panel</a>
       <a href="https://forms.grandsvc.mesh.nycmesh.net/">Other Forms</a>
     </div>
+    <div style="display:flex; flex-direction: row; justify-content: space-between; width:700px; margin-top: 20px">
+      <a href="/api/v1/">MeshDB Data API</a>
+      <a href="/api-docs/swagger/">API Documentation (Swagger)</a>
+      <a href="/api-docs/redoc/">API Documentation (Redoc)</a>
+    </div>
```
WillNilges, line 108: Can we change `api-docs` to `docs`?

WillNilges (changes requested):
Thanks for putting this together. Mostly just wording of docstrings, plus a few organizational concerns.

WillNilges:
Thoughts on adding a link to https://wiki.mesh.nycmesh.net/books/services-software/page/meshdb-schema-design

src/meshdb/templates/rest_framework/login_base.html
Andrew-Dickinson: Credentials are only necessary to use the interactive "call this endpoint now" features of `swagger-ui`. They are not necessary to view the endpoint & schema descriptions, or to query any public endpoints

src/meshapi/urls.py
Andrew-Dickinson: You already approved this in a separate PR lol: https://github.com/WillNilges/meshdb/pull/158

src/meshapi/models.py
Andrew-Dickinson: I checked the existing map, confirmed this is `km`, I will fix this note

src/meshapi/urls.py
Andrew-Dickinson: The whole `/mapdata` path is purely for backwards compatibility. I renamed it to `nodes` to more accurately reflect the hacky "installs get overwritten with NNs" behavior. It doesn't just show you installs, it shows you `nodes` with all of the historical baggage that term brings

Andrew-Dickinson: But heard, I will add a disclaimer on each endpoint

Andrew-Dickinson: Screenshot 2024-03-02 at 13 01 59
This feels... excessive to the point of being disrespectful. But if this is what you want we can do it

src/meshapi/views/forms.py
Andrew-Dickinson: I don't think either of our summaries is accurate to what we are actually doing in the logic below. It looks to me like we always create the Install object, even if this member already has one on this building. If we want to change this, that work should be captured in a separate issue. For now I'm going to simplify this text to more accurately reflect the current logic:

`"Request received, an install has been created (along with member and building objects if necessary)."`

src/meshapi/views/forms.py
Andrew-Dickinson: I'm nervous to link stuff since our URLs are almost certain to change in the future

src/meshapi/views/forms.py
Andrew-Dickinson: Good catch, fixed

src/meshdb/settings.py
Andrew-Dickinson: Done, but I don't feel great about it. See above

src/meshdb/urls.py
Andrew-Dickinson: These are just API docs and they will only ever be API docs, we can't add other stuff to the page without some template pain (and this should probably live on the wiki anyway). I stand by `api-docs`

Andrew-Dickinson:
> Thoughts on adding a link to https://wiki.mesh.nycmesh.net/books/services-software/page/meshdb-schema-design

It's probably worth adding a high level "meshdb docs" wiki page. Maybe we should link that instead?

src/meshdb/settings.py
Andrew-Dickinson: I don't know that caps screaming at folks makes as much sense here. We didn't maintain schema compatibility with the existing endpoints (since we don't render CSVs). If we didn't want anyone to integrate against this, why did we build it?

src/meshdb/settings.py
Andrew-Dickinson: Our docs on this are quite thin but sure, linked this page https://docs.nycmesh.net/hardware/liteap/

src/meshapi/views/lookups.py
Andrew-Dickinson: I have no idea what this means

src/meshapi/views/geography.py
Andrew-Dickinson: Actually they're so that we can use `IgnoreClientContentNegotiation` since we were throwing exceptions when clients requested the (correct) content type of `application/vnd.google-earth.kml+xml`

src/meshapi/urls.py
WillNilges: OK then, perhaps just a 
```
(Warning: This endpoint is a legacy format and may be deprecated/removed in the future)
```

src/meshapi/urls.py
WillNilges: oh pfft

src/meshapi/views/forms.py
WillNilges: Nah, let's keep it. If someone needs to work on the join form (For instance if they're working on the wizard onboarder idea) then they would need it.

src/meshapi/views/forms.py
WillNilges: Oh, I guess you're right.

https://github.com/WillNilges/meshdb/blob/main/src/meshapi/views/forms.py#L171

Yeah I don't think that lines up with what we wanted 😂

src/meshapi/views/geography.py
WillNilges: wtf, odd...

src/meshdb/settings.py
WillNilges: Well, I see the Query Form that I built as a hack, that's all. I would like to re-think the way that all of our current tools work, and I want people to build something that uses the "base" API, where you're interacting with the actual model objects.

src/meshdb/urls.py
WillNilges: Well, I wasn't thinking of adding "other stuff" to these docs, it's just that these are _the_ docs for MeshDB. I see where you're coming from though. IDRC.

WillNilges:
> > Thoughts on adding a link to https://wiki.mesh.nycmesh.net/books/services-software/page/meshdb-schema-design
> 
> It's probably worth adding a high level "meshdb docs" wiki page. Maybe we should link that instead?

Definitely

WillNilges (approved):

PR #215, opened 2024-02-24
Replace original in the custom inline template

WillNilges:
Seems like removing this from the custom template I made broke editing objects in the `change` menu, so I put it back. Wondering if there's a way to make tests for this kind of thing now.


WillNilges:
https://www.selenium.dev/selenium-ide/

I think we'll need something like Selenium to properly test this.

WillNilges:
This might need further investigation. I will make an issue. I wanna get this PR out the door because it at least fixes the issue.

Andrew-Dickinson (approved):

src/meshapi/tests/test_join_form.py
```diff
@@ -322,7 +322,7 @@ def test_member_moved_and_used_stripe_email_join_form(self):
         # and we will connect it to their old registration
         v_sub_3 = valid_join_form_submission.copy()
         v_sub_3["email"] = "jsmith+other@gmail.com"
-        v_sub_3["street_address"] = "153 Broome Street"
+        v_sub_3["street_address"] = "178 Broome Street"
```
WillNilges, line 325: I had to change this because our API was consistently giving us 1000000 (a bad BIN) on 153 broome st. I guess they yeeted that BIN, but it's still working on NYC DOB website?
https://a810-bisweb.nyc.gov/bisweb/PropertyProfileOverviewServlet?boro=1&houseno=153&street=Broome+St&go2=+GO+&requestid=0

Not really sure what's going on.

WillNilges: The test works with 178 Broome St

PR #216, opened 2024-02-24
Add support for grabbing, listing, and displaying panoramas

### LINKED ISSUE #173: Integrate with nodedb to pull links to panoramas?
https://github.com/nycmeshnet/node-db/tree/master

### LINKED ISSUE #214: Better ArrrayFields for things like emails and panorama links
Working on #173 and I am having trouble showing links to panos effectively
image

This also affects emails

image

I found this that might work for the email field

https://pypi.org/project/django-better-admin-arrayfield/

But I want to find some nice way to display the panos in the admin panel.

WillNilges:
Panorama support! This PR accomplishes (or will accomplish) several things:
- [x] Downloading panorama list from [GitHub](https://github.com/nycmeshnet/node-db/tree/master/data/panoramas)
- [x] Adds panorama URLs as a field in Buildings
- [x] Listing panoramas for each building in the admin panel
- [ ] ~Displaying a panorama as a thumbnail in the building list in the admin panel~
    - _This would probably require shrinking and caching the panoramas server-side, and that's probably super overkill for this. Might make an issue for it later. We'll see._
- [x] List panoramas in the Admin Panel
    - [x] Allow users to view panoramas in/from the Admin Panel
- [x] Serving Panorama links for the Map endpoint
- [x] Add GitHub Token
- [x] Security
- [x] Locks?

### Tests
- [x] Basic file parsing
- [x] GitHub Tree API Integration
- [x] Adding panorama to a Building

Closes #173

Closes #214


WillNilges:
We should make all these massive map routes return json, instead of html. It's a LOT of HTML.

WillNilges:
Maybe I should write tests for the lock and the security.... hmmmm

WillNilges:
I did a basic security one, but I am very tired now. Maybe will do a race condition one later, basically try to edit photos while an update panorama request is going and see what happens.

src/meshapi/models.py
```diff
@@ -46,12 +47,17 @@ def __str__(self):
...
+            return self.panoramas[0]
+        return None
+
 
 class Member(models.Model):
     name = models.TextField()
     primary_email_address = models.EmailField(null=True)
     stripe_email_address = models.EmailField(null=True, blank=True, default=None)
-    additional_email_addresses = ArrayField(EmailField(), null=True, blank=True, default=list)
+    additional_email_addresses = JSONFormArrayField(models.EmailField(), null=True, blank=True, default=list)
```
Andrew-Dickinson, line 60: What are the implications of this? Just that it's stored as a JSON array internally?

Or this makes the admin display nicer?

src/meshapi/serializers/map.py
```diff
@@ -73,7 +75,9 @@ class Meta:
     installDate = JavascriptDateField(source="install_date")
     roofAccess = serializers.BooleanField(source="roof_access")
     notes = serializers.SerializerMethodField("get_start_of_notes")
-    panoramas = serializers.ReadOnlyField(default=[])  # FIXME: THIS WILL REMOVE ALL PANORAMAS FROM THE MAP UI
+    panoramas = serializers.SerializerMethodField(
+        "get_panorama_filename"
+    )  # FIXME: THIS WILL REMOVE ALL PANORAMAS FROM THE MAP UI
```
Andrew-Dickinson, line 80:
```suggestion
    )
```

src/meshapi/urls.py
```diff
@@ -28,6 +28,7 @@
     path("mapdata/links/", views.MapDataLinkList.as_view(), name="meshapi-v1-map-data-links"),
     path("mapdata/sectors/", views.MapDataSectorList.as_view(), name="meshapi-v1-map-data-sectors"),
     path("geography/whole-mesh.kml", views.map_kml, name="meshapi-v1-geography-whole-mesh-kml"),
+    path("update-panoramas/", views.update_panoramas_from_github, name="meshapi-v1-update-panoramas"),
```
Andrew-Dickinson, line 31: Hm maybe we should do this as an async background task instead of exposing it as an HTTP endpoint?

Andrew-Dickinson: Is there any reason we'd need to do this on-demand?

src/meshapi/widgets.py
```diff
@@ -0,0 +1,34 @@
...
+        value_as_array = json.loads(value)
+        return {
+            "widget": {
+                "name": name,
+                "value": value_as_array,
+            }
+        }
+
+    def render(self, name, value, attrs=None, renderer=None):
+        # Render the JSONFormWidget to allow editing of the panoramas
```
Andrew-Dickinson, line 26: What kind of editing to do we allow? How do we ensure that edits are not "blown away" by the next import

src/meshapi/tests/test_update_panos_github.py
```diff
@@ -0,0 +1,136 @@
...
+    def test_update_panoramas_unauthenticated(self):
+        response = self.c.get("/api/v1/update-panoramas/")
+        code = 403
+        self.assertEqual(
+            code,
+            response.status_code,
+            f"status code incorrect. Should be {code}, but got {response.status_code}",
+        )
+
+    # This tests the endpoint, but not the actual full pipeline.
```
Andrew-Dickinson, line 66: Why doesn't this test the full pipeline?

src/meshapi/views/panoramas.py
```diff
@@ -0,0 +1,170 @@
...
+    return master["commit"]["commit"]["tree"]["sha"]
+
+
+# Returns all the filenames, stripped of extensions and everything
+def list_files_in_git_directory(owner: str, repo: str, directory: str, tree):
+    url = f"https://api.github.com/repos/{owner}/{repo}/git/trees/{tree}?recursive=1"
+    token = os.environ.get("PANO_GITHUB_TOKEN")
+    response = requests.get(url, headers={"Authorization": f"Bearer {token}"})
+    if response.status_code != 200:
+        print(f"Error: Failed to fetch GitHub directory contents. Status code: {response.status_code}")
```
Andrew-Dickinson, line 163: Should this be an exception?

src/meshapi/views/panoramas.py
```diff
@@ -0,0 +1,170 @@
...
+            panos[number] = [f]
+        else:
+            panos[number].append(f)
+    return panos
+
+
+# This is awful. Maybe there are easy ways to generalize some cases like stripping
+# spaces, but for now I would rather explicitly handle these cases until I have
+# better tests.
+def parse_pano_title(title: str):
```
Andrew-Dickinson, line 110: Wow this is literally like a leetcode problem I'm so sorry

src/meshapi/views/panoramas.py
```diff
@@ -0,0 +1,170 @@
...
+                    f"Warning: Could not add panorama to building (Install #{install_number}). Install does not exist."
+                )
+                warnings.append(install_number)
+                continue
+            for filename in filenames:
+                file_url = f"{host_url}{filename}"
+                panoramas.append(file_url)
+            if install.building.panoramas == panoramas:
+                continue
+            install.building.panoramas = panoramas
```
Andrew-Dickinson, line 83: This overwrites earlier installs with the panos from later ones, which is not ideal, but is understandable. Could you just add a comment pointing this out and saying maybe we should fix this in the future?

src/meshapi/urls.py
Andrew-Dickinson: We could also just schedule it to run hourly with celery

src/meshapi/widgets.py
WillNilges: Hmm, good point. I think I need to update the import logic to account for that. Can probably just check if the order/length is different from expected and instead of replacing the field, append it.

src/meshapi/models.py
Andrew-Dickinson: Right, but why did you change the email addresses?

src/meshapi/widgets.py
Andrew-Dickinson: Alternatively, Olivier is comfortable working with the current system, we could keep ours read-only until we can replace the GitHub

src/meshapi/tests/test_update_panos_github.py
WillNilges: I'm lazy, and when I wrote that test I hadn't yet broken the code up to be easily unit tested yet. I think this is now a relatively atomic test. Testing the whole thing front-to-back is probably unnecessary now.

src/meshapi/views/panoramas.py
WillNilges: That's what I was thinking the whole time.

src/meshapi/views/panoramas.py
WillNilges: Oh, I guess we can drop data. Per my earlier reply, I'll see if I can create some logic to protect diffs and do something sensible.

Andrew-Dickinson: Well the cases are slightly different. Here, even if we start with a clean DB, this will cause earlier installs for the same building to get overwritten by later ones

src/meshapi/models.py
WillNilges: so that it also gets the better widget.

src/meshapi/widgets.py
WillNilges: I will add an env var to remove the edit field

PR #223, opened 2024-03-14
Implement the new schema

Andrew-Dickinson:
Migrates to the new schema design. Primary changes are the addition of `Node` and `Device` tables

This forced an overhaul of nearly everything. That overhaul is incomplete, here's an (incomplete) task summary:
- [x] Schema changes
- [x] Get tests passing
- [ ] Admin UI overhaul
- [ ] Import script overhaul (see #224 )
- [ ] Fix merge conflicts with #216 
- [ ] Address-based de-duplication in join form
- [ ] KML & Map correctness checks
- [ ] Query & Lookup forms need some love (and tests)

PR #225, opened 2024-03-16
Update import script to account for new schema

Andrew-Dickinson:
Builds on #223, implements import script changes

Most of the way there, but will probably need a follow-up PR to tweak some edge cases


WillNilges (approved):
I trust you

PR #235, opened 2024-03-17
Update Admin Panel for new Schema

WillNilges:


WillNilges:
Selecting a device on the link edit page is impractical (some kind of search is needed)
Could we use the name of the node instead of the node number on the list view?
Can we list the links on the node table?
On the node page can we list sectors by name instead of SSID? (NN2090 looks a bit strange cus of the wds)
Can we show buildings on the node page?
The node editor on the building page is a bit clunky, for example you have to type Vernon not 5916 . Any way to improve the search fidelity? Or maybe just include the NN in every row (e.g. Vernon (NN5916))

WillNilges:
haha fuck: https://code.djangoproject.com/ticket/9025

Grandchildren inlines aren't supported, but there is a package for it.

WillNilges:
Not sure if this does what we want: https://github.com/OskarPersson/django-nested-inline

Andrew-Dickinson (approved):
Looks great!

PR #247, opened 2024-03-21
Implement lookup endpoints for the new schema using django-filter

Andrew-Dickinson:
Uses `django-filter` to implement lookup endpoints based on the new schema. Also adds comprehensive test coverage for the available filters.

Query form endpoints still need to be fixed, but using this pattern should make that pretty straightforward


src/meshapi/views/lookups.py
```diff
@@ -1,12 +1,61 @@
...
+            return super().get(request, *args, **kwargs)
+
+    def get_queryset(self):
+        result = super().get_queryset()
+
+        # This is a search endpoint, if the user doesn't supply any filters,
+        # they expect that we return nothing instead of everything.
+        # If they wanted everything they'd use the appropriate list endpoint
+        if not self.request.query_params:
+            return result.none()
```
WillNilges, line 39: Maybe we should return a 400 instead of nothing?

WillNilges (approved):

src/meshapi/views/lookups.py
Andrew-Dickinson: Done

PR #271, opened 2024-03-24
Downgrade to Postgres 15.6

WillNilges:
~Using django-dbbackup and celery, take automated backups of the database.~ Fix version skew issues with postgres and use 15.6 everywhere, pinning the dockerfiles.

- [x] Downgrade Postgres 16 -> 15
- [ ] ~Configure django-dbbackup~
- [ ] ~Configure automated backups~
- [ ] ~Verify backups exist in S3~


WillNilges:
Downgrading postgres will require another database import.

Andrew-Dickinson (approved):

PR #278, opened 2024-03-26
Setup .env in test run

### LINKED ISSUE #276: Fix flaky panoramas test
https://github.com/WillNilges/meshdb/pull/273

The tests on this PR should not be failing, it's a readme change lol

We need to mark that test as flaky, mock the GitHub API call, or remove it

james-otten:
Resolves #276

I ran into very similar issues with unit tests not accepting env var values for the `PANO_*` variables until I set them in my `.env` file. This PR tries to solve #276 in a similar way to how I solved this on my local machine.

I think it is due to the `dotenv` call in `src/manage.py`, but I'm not 100% sure.

Edit: I tested on my fork, it looks like it works!  https://github.com/james-otten/meshdb/pull/1

### DIFF
--- a/.github/workflows/run_django_tests.yaml
+++ b/.github/workflows/run_django_tests.yaml
@@ -40,21 +40,23 @@ jobs:
         run: "pip install --upgrade pip"
       - name: "Install package"
         run: pip install ".[dev]"
+      - name: "Setup .env"
+        run: |
+          echo "\n" >> .env
+          echo "DB_NAME='nycmesh-dev'" >> .env
+          echo "DB_USER='nycmesh'" >> .env
+          echo "DB_PASSWORD='abcd1234'" >> .env
+          echo "DB_HOST='localhost'" >> .env
+          echo "DB_PORT='5432'" >> .env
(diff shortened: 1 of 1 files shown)


WillNilges:
Ah, looks like there were problems loading the panorama env vars still 😅 
https://productionresultssa1.blob.core.windows.net/actions-results/e1257899-d531-45f7-ac1e-c95cac421864/workflow-job-run-152bb3b5-7908-51f7-367b-02ffebb7d905/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-03-27T07%3A39%3A59Z&sig=2%2F2aKa3qI3%2BA1jFp3%2BjxqAGl8qJD2O0gVdGhZv01SaA%3D&sp=r&spr=https&sr=b&st=2024-03-27T07%3A29%3A54Z&sv=2021-12-02

I might need to revisit the way I'm pulling in these env vars. It's also like... an unnecessary amount of vars.

I also think we should mock the Github endpoint, because it has given me a 500 in a variety of contexts and I think is, simply put, flaky.

james-otten:
Use #285 instead

PR #289, opened 2024-03-29
Add IP addresses to devices

### LINKED ISSUE #243: Track IP address(es) for devices
Import IP addresses from UISP and make them clickable in the admin UI

Andrew-Dickinson:
Adds IP addresses to devices, automatically imported from UISP, and displayed on the admin UI, with a clickable link to navigate to the displayed IP. Closes #243


WillNilges (commented):
Nice, thanks

What about invalid/broken IP addresses? It would be really cool if you could make it disable the view button until it was able to ping or something.

Also,what if when you click "view", it replaces the map view, or opens a tab or something? Just to avoid breaking _the flow™_

Andrew-Dickinson:
I switched the IP address field type to `GenericIPAddressField`, which should prevent invalid addresses. Of course addresses could be un-routable, etc. But since that could be quite transient it seems like not our place to validate

Opening in a new tab is what it already does, I think this is sufficient for now

WillNilges (approved):

PR #290, opened 2024-03-29
Setup automated backups with django-dbbackup

WillNilges:
Installs `django-dbbackup` to provide the ability to back up our database automatically. Configures a celery beat to do so hourly.

Also fixes celery beat for syncing panoramas, and tweaks the `entrypoint.sh` to avoid celery locking up and using 100% CPU.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -5,6 +5,12 @@ DB_HOST=localhost
 DB_PORT=5432
 DB_PASSWORD=abcd1234
 
+# Backups
+AWS_ACCESS_KEY_ID=
+AWS_SECRET_ACCESS_KEY=
+BACKUP_S3_BUCKET_NAME=meshdb-data-backups
+BACKUP_S3_BASE_FOLDER=meshdb-backups/development/
+
 CELERY_BROKER=redis://localhost:6379/0
 
 # DO NOT USE THIS KEY IN PRODUCTION
--- a/src/meshapi/views/panoramas.py
+++ b/src/meshapi/views/panoramas.py
@@ -2,7 +2,6 @@ import os
 from pathlib import Path
 
 import requests
-from celery.schedules import crontab
 from rest_framework.decorators import api_view, permission_classes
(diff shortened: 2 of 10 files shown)


src/meshapi/views/panoramas.py
```diff
@@ -19,25 +19,38 @@
...
+                "detail": f"Saved {panoramas_saved} panoramas. Got {len(warnings)} warnings.",
+                "saved": panoramas_saved,
+                "warnings": len(warnings),
+                "warn_install_nums": warnings,
+            },
+            status=status.HTTP_200_OK,
+        )
+    except (ValueError, GitHubError) as e:
+        print(e)
+        return Response({"detail": e}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
Andrew-Dickinson, line 49: Is this allowed?
```python
try:
    bad = 1/0
except Exception as e:
    result = {"a": e}
    
import json
json.dumps(result)
```
Gives me the following error:
```
TypeError: Object of type ZeroDivisionError is not JSON serializable
```

Andrew-Dickinson (approved):

.env.sample
```diff
@@ -5,6 +5,12 @@ DB_HOST=localhost
 DB_PORT=5432
 DB_PASSWORD=abcd1234
 
+# Backups
+AWS_ACCESS_KEY_ID=
+AWS_SECRET_ACCESS_KEY=
+BACKUP_S3_BUCKET_NAME=meshdb-data-backups
+BACKUP_S3_BASE_FOLDER=meshdb-backups/development/
+
```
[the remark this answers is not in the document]
WillNilges: I should test that. I think so, since each task should be handled atomically, but might need to add some try/excepts in there.

src/meshapi/views/panoramas.py
WillNilges: oops

PR #292, opened 2024-03-30
Add type field to the node model

### LINKED ISSUE #178: Make map hub/AP display not based on notes field
Currently we proxy along the "notes" field from the spreadsheet by string splitting the imported value from our combined one. We should do the string parsing at import time and add a separate DB field to indicate the node status as a hub/AP. 

We can then re-build the note that the map javascript expects to see as a string based on the information in the DB fields. This is important because if an admin edits the notes formatting they could break the map display

Andrew-Dickinson:
Adds a new field to the node model which controls how a given node appears on the map. I'm flexible on naming here if we think there's a better way to describe this concept.

Currently, we just parse the "notes" that we import from the spreadsheet, which is dangerous (can leak private info) and brittle (will break if an admin changes the notes). This PR refactors that so that the parsing for node "type" identification happens at import time, and then we re-create "synthetic" notes to communicate back to the frontend in a format it expects

Closes #178


src/meshapi/models/node.py
```diff
@@ -26,6 +34,12 @@ class NodeStatus(models.TextChoices):
 
     status = models.CharField(choices=NodeStatus.choices, help_text="The current status of this Node")
 
+    map_display = models.CharField(
```
WillNilges, line 37: Perhaps this could be named a little more generically, as this might be useful in other contexts besides the map. Perhaps we could name it something like `variant`, `type`, or `icon`?

WillNilges: `display_variant`, even 😛

Andrew-Dickinson: `type` sounds good to me

WillNilges (approved):

Issue #88, opened 2023-11-05
Add basic monitoring


Andrew-Dickinson:
We should receive some kind of basic alerts for `5xx` errors in production. Probably a slack message?

vivinano:
I created some middleware and testing it out now possible super basic but seems to do error handling

Andrew-Dickinson:
We should also include monitoring of the "supplemental" services. Admin map, regular map, LOS DB, etc

Andrew-Dickinson:
New requirement: if the celery workers haven't run in a long time we should be alerted

PR #300, opened 2024-03-31
Graceful failure for nyc open data

### LINKED ISSUE #227: Fail gracefully when the NYC opendata site goes down
https://data.cityofnewyork.us/ apparently goes offline sometimes. We need access to https://data.cityofnewyork.us/resource/qb5r-6dgf.json in order to get building altitude at join form submission time. However when this API is offline we should continue to accept join-form submissions, we can always backfill later

james-otten:
- Resolve #227
    - Only make one attempt to fetch altitude data from nyc open data
    - If altitude cannot be determined for any reason, a default value of `-1` is used to signal that it should be corrected manually
- Add a default HTTP [timeout](https://docs.python-requests.org/en/latest/user/advanced/#timeouts) of 5 seconds for both external HTTP requests during address validation
- Switch from `print` to `logging.*` in touched files
- `isort` and `black` on touched files


src/meshapi/validation.py
```diff
@@ -61,10 +62,14 @@ def __init__(self, street_address: str, city: str, state: str, zip: int):
...
-            nyc_planning_req = requests.get(f"https://geosearch.planninglabs.nyc/v2/search", params=query_params)
+            nyc_planning_req = requests.get(
+                "https://geosearch.planninglabs.nyc/v2/search",
+                params=query_params,
+                timeout=DEFAULT_EXTERNAL_API_TIMEOUT_SECONDS,
+            )
             nyc_planning_resp = json.loads(nyc_planning_req.content.decode("utf-8"))
         except Exception as e:
-            print(f"Got exception querying geosearch.planninglabs.nyc: {e}")
+            logging.error(f"Got exception querying geosearch.planninglabs.nyc: {e}")
```
Andrew-Dickinson, line 72: nit: You can actually just do this. The logger can automagically get the exception from the context 
```suggestion
            logging.exception(f"Got exception querying geosearch.planninglabs.nyc")
```

src/meshapi/validation.py
```diff
@@ -102,17 +107,29 @@ def __init__(self, street_address: str, city: str, state: str, zip: int):
...
+                self.altitude = float(nyc_dataset_resp[0]["heightroof"]) + float(nyc_dataset_resp[0]["groundelev"])
+        except OpenDataAPIError:
+            self.altitude = INVALID_ALTITUDE
+            logging.warning(
                 f"(NYC) DOB BIN ({self.bin}) not found in NYC OpenData while trying to query for altitude information"
             )
-
-        self.altitude = float(nyc_dataset_resp[0]["heightroof"]) + float(nyc_dataset_resp[0]["groundelev"])
+        except Exception as e:
+            self.altitude = INVALID_ALTITUDE
```
Andrew-Dickinson, line 134: Altitude is optional on the `Building` model, so we maybe should make this `None` here? I think that would be better than having to check for `None` _and_ `-1` everywhere that reads altitude values. Also, technically `-1` could be a valid altitude (for an extremely short building in an extremely flood prone area).

src/meshapi/validation.py
```diff
@@ -61,10 +62,14 @@ def __init__(self, street_address: str, city: str, state: str, zip: int):
                 "text": self.address,
                 "size": 1,
             }
-            nyc_planning_req = requests.get(f"https://geosearch.planninglabs.nyc/v2/search", params=query_params)
+            nyc_planning_req = requests.get(
+                "https://geosearch.planninglabs.nyc/v2/search",
+                params=query_params,
+                timeout=DEFAULT_EXTERNAL_API_TIMEOUT_SECONDS,
+            )
```
Andrew-Dickinson, line 69: I know that #227 only covers the altitude endpoint, but I'm wondering if there's a way we can be resilient to the failure of this endpoint as well?

Maybe this is a larger concern that should be covered in a separate issue, but it would be nice if we could fall back to [our local](https://github.com/nycmeshnet/meshdb/blob/main/.env.sample#L13) [pelias](https://github.com/pelias/parser) container in these cases

Andrew-Dickinson (commented):
This looks great James, any chance we can get an additional join form test which mocks `request.get` to simulate the failure and confirm the appropriate behavior?

src/meshapi/validation.py
Andrew-Dickinson: If you're going to look into pelias parsing, take a look at the import script which uses pelias extensively. You can probably re-use a lot of the logic

src/meshapi/validation.py
james-otten: Good call, changed the constant.

src/meshapi/validation.py
james-otten: I agree, was going to split out "graceful failure if geosearch.planninglabs.nyc is unavailable", but I can do that here. Also, I split out #304 to fall back to pelias.

james-otten:
> This looks great James, any chance we can get an additional join form test which mocks `request.get` to simulate the failure and confirm the appropriate behavior?

Added some tests to simulate good/invalid/no responses.

src/meshapi/tests/test_validation.py
```diff
@@ -0,0 +1,109 @@
...
+        test_cases = [
+            mock_series_of_tubes,
+            mock_no_list,
+            mock_no_value,
+            mock_null_value,
+            mock_wrong_type,
+            Exception("Pretend this is a network issue"),
+        ]
+
+        for mock_test_case in test_cases:
```
Andrew-Dickinson, line 90: nit: I think if you do this with [`pytest.mark.parametrize`](https://docs.pytest.org/en/7.0.x/how-to/parametrize.html) these tests would be a bit easier to debug/isolate, but don't stress, this is probably fine

Andrew-Dickinson (approved):

Andrew-Dickinson:
Just thought of something real quick, can we add a `TODO` or an issue (maybe a comment on #88?) that we should have an alarm that's triggered whenever this endpoint fails? (so we know that we might need to back-fill some altitude info)

PR #305, opened 2024-03-31
More flake8 stuff

james-otten:
- Add logging of eaten exceptions
- Make imports explicit
- Cleanup unused imports
- Lambda -> nested function
- Some suppressions


src/meshapi/models/__init__.py
```diff
@@ -1,6 +1,6 @@
-from .building import *
```
Andrew-Dickinson, line 1: nit: I think there is a way to do this based on [file path globs](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-per-file-ignores). Something like:

```
per-file-ignores =
    **/__init__.py:F401,F403
```

in `.flake8`

Andrew-Dickinson: Might also want to add `src/meshapi_hooks/models.py`?

Andrew-Dickinson (approved):

PR #306, opened 2024-04-01
Import phone numbers from dupes when we don't already have one

Andrew-Dickinson:
Some small tweaks to the phone number import logic if we already have a member object, but that member object doesn't have a phone number, we should take the new one if it's available


WillNilges (approved):
Assuming you mean that if we happen to come across a duplicate entry in the database (same name or whatever) we grab that phone number? Fine by me.

Andrew-Dickinson:
Only if the we didn’t already have a phone number, but yes

PR #313, opened 2024-04-02
Rename foreign key DB columns from node_id to network_number

Andrew-Dickinson:
While writing a raw SQL query today, I noticed that we have let `node_id` slip into our SQL schema in foreign key columns on the `Building`, `Install`, and `Device` tables. This is not a problem for our Python code or our JSON api payloads, as those use `node` which is imperfect, but much better than `node_id`.

This PR uses django's escape hatch for changing DB column names to change these foreign key columns to `network_number`. 

Unfortunately Django doesn't provide a neat way to do this for many-to-many fields, so I stole [this guy's](https://forum.djangoproject.com/t/add-ability-to-define-custom-column-names-in-the-table-created-by-the-manytomanyfield/25427/6) code for a custom field that makes this easy. (🎶  I hope some day Brent O'Connor rides his f*cking train 🎶 )

This change should be 100% interface safe and have no implications for any python code or JSON payloads (the purpose of the ORM is so we don't have to think about SQL column names anywhere). This is confirmed by the tests passing

I expect this PR to be at least somewhat controversial. Roast me


WillNilges (commented):
Like, I guess this is fine, but I would be curious to know if that guy in that forum post ever made a real PR.

Maybe you should make a real PR :eyes:, become a :sparkles: django contributor :sparkles:

WillNilges (approved):
Approved on the condition that this is easy to revert, which is demonstrably true.

Issue #314, opened 2024-04-02
Fix import for re-used NN rows


Andrew-Dickinson:
Currently we drop any node with an `X-` in the NN column. However, this results in dropping some information, namely `nodeName`, `notes`, and manually tweaked `BIN/lat/lon/alt`. Modify the import script to handle these rows, but only create `Node` objects for them (and not `Install`, `Member`, or `Building` objects as is typical for a spreadsheet row).

This will require setting `status="Planned"` when constructing these `Node` objects and relying on the later logic at `Install` create to change this as needed. 

For these rows, follow the following import guidelines:

Ignored by MeshDB (when status = NN assigned )
- location
- apt
- name
- email
- stripe email
- 2nd email
- phone
- neighborhood
- rooftop access
- referral
- contact notes
- install date
- abandon date
- install notes
- notes2

Imported by MeshDB
- node name -> should be used for the imported node name
- notes -> should be added to imported node notes (and used to determine node type)
- ID -> should be used as the imported NN
- NN -> should be X-, but don't use, we will catch the `Install` -> `Node` relation in the other direction when we get to that install's row
- lat/lon/alt -> should be used for the imported node location
- BIN -> store in a global mapping of `BIN -> NN`, which is iterated at the end of the import process to confirm the appropriate `Node` <-> `Building` relationships are formed

[Slack thread](https://nycmesh.slack.com/archives/C06GW3Q92TY/p1712021238034559?thread_ts=1711769003.400379&cid=C06GW3Q92TY)

PR #315, opened 2024-04-02
Only show node name labels on NN map dots

Andrew-Dickinson:
Based on conversation with Olivier [here](https://nycmesh.slack.com/archives/C06GW3Q92TY/p1711983533597739?thread_ts=1711769003.400379&cid=C06GW3Q92TY), we should only show the node name on the NN dot, not the install dots. 

There's some follow-on work here to properly import node-name in some re-used NN edge cases, so this won't perfectly satisfy Olivier. I've captured that follow-on work via #314


WillNilges (commented):
whoops wrong pr

WillNilges (approved):
God I hate the old map

PR #317, opened 2024-04-02
Make the test postgres images match prod

james-otten:
Make the postgres image tags in CI/CD match the one in [docker-compose.yaml](https://github.com/nycmeshnet/meshdb/blob/main/docker-compose.yaml#L14)


WillNilges (approved):
❤️

PR #319, opened 2024-04-02
Add pgadmin container to docker-compose

### LINKED ISSUE #312: Allow raw SQL queries for analytics access
Come up with a way to allow raw SQL queries (via a read-only user) for analytics workloads. Export to CSV. Ideally this would be via some type of UI, so that folks don't have to download a postgres client. Maybe there's some kind of Django analytics library that helps with this?

https://nycmesh.slack.com/archives/C06GW3Q92TY/p1712013900917419

WillNilges:
Sets up a pgadmin container to allow for easy access to the database for making raw queries. Pgadmin should store configuration such as user accounts in a separate Docker volume on the host. It requires credentials to meshdb's database to be added per-account, so we could create (if we don't already have one) a read-only account on the database and connect peoples' accounts to that.

This should be a tool with individual accounts granted to people (CC @james-otten) by request. There probably aren't that many people who need access to this.

Resolves #312


WillNilges:
Ah shit I really gotta finish this

docker-compose.yaml
```diff
@@ -33,6 +33,14 @@ services:
     image:
       redis
 
+  pgadmin:
+    networks:
+      - api
+    image: dpage/pgadmin4:8.9
+    environment:
+      PGADMIN_DEFAULT_EMAIL: ${PGADMIN_EMAIL}
+      PGADMIN_DEFAULT_PASSWORD: ${PGADMIN_PASSWORD}
```
james-otten, line 42: Will need to be added to grand. This one needs to be a good one.

james-otten (approved):

PR #327, opened 2024-04-04
Convert print -> logging

james-otten:
- Convert `print()` to `logging.*()`, especially `logging.exception()` for exceptions
- In one case, limit error message detail to the class name
- Add a `.gitignore` entry for myself

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -175,3 +175,6 @@ loadenv.sh
 
 # Spreadsheet data
 spreadsheet_data
+
+# vs code
+.devcontainer.json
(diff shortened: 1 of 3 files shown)


WillNilges:
Thanks for taking this on. Thoughts on doing some basic logging config?
```
logger = logging.getLogger(__name__)
logging.basicConfig(filename='example.log', encoding='utf-8', level=logging.DEBUG)
```

Also, I [un-fixed](https://github.com/nycmeshnet/meshdb/commit/afbfd9ec6ec08e07fce42cd077134e0a979609f1) the bug on the `api/v1/mapdata/nodes` to test how it logged errors like that, and it seems that it doesn't. I'd really like that for more quickly debugging production errors. Perhaps a setting in logging, django, gunicorn or some combo of the three? 😅

WillNilges (approved):

PR #334, opened 2024-04-06
Use flickity to display panoramas

WillNilges:
TIL that a component called a carousel exists and there are libraries for it. This uses one of said libraries, "Flickity" to (hopefully) make the panoramas display a little better.


WillNilges:
image
image
image

src/meshdb/settings.py
```diff
@@ -41,6 +41,7 @@
     "db.mesh",
     "db.nycmesh.net",
     "127.0.0.1",
+    "10.0.2.2",
```
Andrew-Dickinson, line 44: Why do we want to allow this?

Andrew-Dickinson (approved):

src/meshapi/templates/widgets/panorama_viewer.html
```diff
@@ -1,5 +1,15 @@
-<div style='display:flex; flex-direction: row; width:100%; overflow-x:auto; max-height:300px; padding-left: 10px;'>
-	{% for panorama in widget.value %}
-		<a href="{{ panorama }}" target="_blank"><img src="{{ panorama }}" style="margin-right:10px"/></a>
-	{% endfor %}
+<script src="https://unpkg.com/flickity@2/dist/flickity.pkgd.min.js"></script>
```
james-otten, line 1: Can this safely use SRI?

src/meshdb/settings.py
WillNilges: oops that's an artifact of my testing on windows.

src/meshapi/templates/widgets/panorama_viewer.html
james-otten, line 1:
```suggestion
<script src="https://unpkg.com/flickity@2/dist/flickity.pkgd.min.js" integrity="sha384-6ma6pGRLjeu6PkX+yOaR0dUGt6OQl9evbv3H2kdLWwIXtUztaQrWUlbRBPVpDH4B" crossorigin="anonymous"></script>
```

WillNilges: I'm not sure. I could probably serve a copy of flickity from our server if that would be better.

PR #341, opened 2024-04-12
Search by notes

### LINKED ISSUE #322: Allow searching notes
We should find a way to allow searching for objects via notes. It would be very handy if I could just search "BAM Distro."

I can just search by date, but this feature might be useful somehow else.

james-otten:
Resolve #322 

Implemented for individual objects (search notes of installs when searching installs)


WillNilges:
Awesome, have you noticed whether it pollutes the search results at all? I imagine this is an excellent way to solve the problem I named in the issue but I want to make sure it doesn't degrade the search experience at all.

WillNilges (approved):

PR #344, opened 2024-04-18
Add new building objects from the BuildingInline

WillNilges:
Enables jumping straight to the Building add view from the Node change view. I believe this solves half of the problem.

Update: I've done it, I've cracked the code. You have to use a "through" Inline. That is done here. In a followup PR I will clean up the Admin panel code and implement that for other fields.


james-otten (approved):
Confirmed the new UI exists. Building creation is a mystery to me.

WillNilges:
image

PR #347, opened 2024-04-19
Fallback to Network Number to import panoramas

WillNilges:
It seems like there may be panoramas named with NN instead of Install # after all.


WillNilges:
ugh I have to fix the tests.

WillNilges:
Oh weird, the tests failed, but now they pass. I didn't do anything.

james-otten (approved):

Issue #338, opened 2024-04-08
Panorama Viewer only works for Admins


WillNilges:
Django seemingly has a hardcoded, undocumented obligation to render data as plaintext when you have view-only permissions.
![image](https://github.com/nycmeshnet/meshdb/assets/42927786/19682b08-ccd5-4d82-a127-62e918fb5f2f)

I need to render the panorama viewer. There may be a workaround.

Further reading:

https://code.djangoproject.com/ticket/30577
https://stackoverflow.com/questions/14832739/django-admin-how-to-display-widget-on-readonly-field

WillNilges:
This is still an issue as of 2 months ago.

https://www.reddit.com/r/django/comments/1al8qlv/admin_readonly_field_with_custom_widgets_are/

Andrew-Dickinson:
> I generally find that when I start fighting with Django admin I'm pushing on a closed door and just write a discrete app instead.

An ominous point that I fear we may be bumping into in more ways than one...

PR #353, opened 2024-04-22
Display panoramas in Node

WillNilges:
This is such a godawful hack. There has to be a better way to do this, but also there's nothing better and more permanent than a "temporary solution"

image


WillNilges:
This does actually enable Installers to see panoramas via the Node view. Decent workaround for #338 ?

james-otten (approved):
New UI is shown

PR #354, opened 2024-04-23
Add codeql scanning

james-otten:
Add basic code security scanning.

Runs on:
- main
- PRs to main
- On a schedule (Saturday mornings)


.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,58 @@
...
+      run: |
+        echo 'If you are using a "manual" build mode for one or more of the' \
+          'languages you are analyzing, replace this with the commands to build' \
+          'your code, for example:'
+        echo '  make bootstrap'
+        echo '  make release'
+        exit 1
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v3
```
WillNilges, line 56: Is it possible to pin these versions (both here and above) to a hash like we do for our other workflows?

WillNilges (approved):

.github/workflows/codeql.yaml
james-otten: Good call

PR #359, opened 2024-04-27
#88 trying out basic middleware implementation for error handling

vivinano:
This is just a basic implmentation. but based on what the task list was saying you guys just wanted something to happen when response code is X so currently it just prints on the console

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ When you're done, you can stop the server with `Ctrl+C`, and run `docker compose
 > [!NOTE]
 > To spin things back up again later, just run:
 > ```sh
+> source .venv/bin/activate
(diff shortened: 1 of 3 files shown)


WillNilges:
Thanks for looking at this! If the PR is not finished, you should be able to "convert to draft" in the top right corner.

vivinano:
> Thanks for looking at this! If the PR is not finished, you should be able to "convert to draft" in the top right corner. Also, do you mind mentioning what issue this relates to?
this was to the #88 
i saw that the first thing that was mentioned was if status code 5xx then post slack message so was figuring that out and create a small snippet of code and was testing it with print statments

Andrew-Dickinson:
Closing in favor of #481 (and others)

PR #363, opened 2024-04-28
Validate full domain

james-otten:
Silence codeql alert by doing further validation of a url.

- https://github.com/nycmeshnet/meshdb/security/code-scanning/9
- https://github.com/nycmeshnet/meshdb/security/code-scanning/10

I do not classify either of these as security defects.


WillNilges (approved):
Seems legit. We appear to only be getting maps.googleapis.com according to the inspector.

image

Issue #266, opened 2024-03-23
More code linters


Andrew-Dickinson:
- [x] MyPy
- [x] Flake8
- [x] isort

james-otten:
Maybe exclude import script -A

Andrew-Dickinson:
Completed via #381

PR #367, opened 2024-05-03
Resolve some mypy findings

james-otten:
Related to #266


Andrew-Dickinson (approved):

Andrew-Dickinson:
You probably know this, but `Any` is kind of a cheat which basically disables `mypy` for those variables. It's fine for scrappy stuff like support scripts, or when the types get stupid/gnarly, but we should try not to use it when it can be avoided

PR #368, opened 2024-05-11
Update terraform and stuff

WillNilges:


WillNilges:
We really oughta look into using this: https://registry.terraform.io/modules/xunleii/k3s/module/latest

WillNilges:
unfortunately I think this will still end up needing ssh and therefore the ssh agent running, which really sucks. At the very least maybe I can stop it from needing the agent.

WillNilges:
I haven't tried these instructions fully, but everything up to step 4 should work. It is super late.

TODO:
- [ ] Solidify networking
- [ ] Solidify database
- [ ] Investigate K3s module for terraform
- [ ] Figure out how to update the IP address of the kube manifest or use DNS
- [ ] uhhhhhhh

WillNilges:
OK so installing kube with Terraform? Hot.

Installing custom software with Terraform? Maybe Not.

It adds another layer of abstraction, another moving part to deal with. I think that managing MeshDB through Helm (or perhaps some other software) directly might work better. I need to play with it. Once I have the kubeconfig, I'm chilling.

I unfortunately need to go to bed I'm so tired.

WillNilges:
OK, the internet seems to be indicating that managing packages with Terraform probably isn't a good idea. I am tempted to bring in Ansible, and move the k3s stuff that we know works OK into its own directory.

infra/helm/meshdb/templates/namespace.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    kubernetes.io/metadata.name: meshdb
+  name: meshdb
```
james-otten, line 6: `.Values.meshdb_app_namespace`

Handled at deployment

james-otten:
> OK, the internet seems to be indicating that managing packages with Terraform probably isn't a good idea. I am tempted to bring in Ansible, and move the k3s stuff that we know works OK into its own directory.

We can instead move the packages to the VM images (like qemu guest agent)

PR #371, opened 2024-05-18
Fix DB import script

Andrew-Dickinson:
The node table no longer has an auto-incrementing sequence, so the import script fails. This PR removes the line that tries to update the auto-increment counter at import time


james-otten (approved):
I approve of that line not being there

PR #379, opened 2024-06-01
Add additional_phone_numbers field to Member object

### LINKED ISSUE #376: Add a way to store multiple phone numbers per member
Add an `additional_phone_numbers` field which functions similar to `additional_emails`. Maybe also add a convenience field for `all_phone_numbers` similar to what we have for emails

https://nycmesh.slack.com/archives/C06GW3Q92TY/p1717271565403239

Andrew-Dickinson:
Closes #376 by adding a field to store additional phone numbers for each member

Stole basically all of the logic from the similar feature we already support for email addresses


src/meshapi/management/commands/scramble_members.py
```diff
@@ -45,6 +45,7 @@ def handle(self, *args: Any, **options: Any) -> None:
                 member.stripe_email_address = ""
                 member.additional_email_addresses = []
                 member.phone_number = fake.phone_number()
+                member.additional_phone_numbers = []
```
WillNilges, line 48: Might be good to generate a few randomly.

WillNilges (approved):
🚢

src/meshapi/management/commands/scramble_members.py
Andrew-Dickinson: Done

Issue #311, opened 2024-04-01
Confirm that primary nodes are elected correctly for all Buildings with multiple nodes to pick from


Andrew-Dickinson:
After #310 merges, run these commands and check each building that is output to make sure the correct node was elected:

```python
from django.db.models import Count
from meshapi.models import Building

multi_node_buildings = list(Building.objects.annotate(num_nodes=Count('nodes')).filter(num_nodes__gt=1).all())
for building in multi_node_buildings:
    print(building)
    print(building.primary_node)
    print(building.nodes.all())
    print("--------------------------------")
```

Andrew-Dickinson:
Spoiler alert, I think it will still not work for some edge cases, NN265/NN511 being one of them. Fix that lol

Andrew-Dickinson:
The following clusters are still wrong:

> 725 4th Avenue, Brooklyn, NY
> NN393
> <QuerySet [<Node: NN393>, <Node: NN5989>]>

Should be 5989

> 36 Monroe Street, New York, NY
> NN994
> <QuerySet [<Node: NN994>, <Node: NN1275>]>

Should be 1275

> 119 Avenue B, New York, NY
> NN338
> <QuerySet [<Node: NN338>, <Node: NN339>]>

Something weird is happening here, looks like a second NN for the same building but the `X-` pointer was wrong. Fixed it, we'll see if this clears things up

> 47 Thames Street, Brooklyn, NY
> NN511 (Thames-2)
> <QuerySet [<Node: NN511 (Thames-2)>, <Node: NN265 (Thames)>]>

Should be 265

Plus grand st is kinda messed up, but I think that should be fixed by some manual fixes I made

Andrew-Dickinson:
This appears to be correct now, closing

PR #382, opened 2024-06-02
Fix: low number nodes are elected primary instead of older nodes

Andrew-Dickinson:
Progress towards #311, may need an additional follow up to fully resolve

### DIFF
--- a/src/meshdb/utils/spreadsheet_import/parse_node.py
+++ b/src/meshdb/utils/spreadsheet_import/parse_node.py
@@ -89,12 +89,11 @@ def normalize_building_node_links(building: models.Building, node: models.Node):
     Scan the node <-> building relations for the given node and building to make sure all buildings
      in the cluster refer to all nodes in the cluster and vice versa
 
-     Also sets the correct primary_node pointer for each Building, selecting any existing primary
-     nodes in the cluster first, and then falling back to the node at hand
+     Also sets the correct primary_node pointer for each Building, by electing the oldest active node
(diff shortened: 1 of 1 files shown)


src/meshdb/utils/spreadsheet_import/parse_node.py
```diff
@@ -106,24 +105,19 @@ def normalize_building_node_links(building: models.Building, node: models.Node):
...
-                logging.error(
-                    f"Detected multiple primary nodes in the same cluster: "
-                    f"{established_primary_nodes}. These will be consolidated "
-                    f"to the oldest node: {established_primary_nodes[0]}."
-                )
-            b.primary_node = established_primary_nodes[0]
-        else:
-            b.primary_node = node
-
+        b.primary_node = primary_node
```
WillNilges, line 122: Nit: Why not just set `b.primary_node` directly above on line 110?

WillNilges (approved):

src/meshdb/utils/spreadsheet_import/parse_node.py
Andrew-Dickinson: Because we have to set it for each building in `buildings_in_cluster`? I'm not sure what you're getting at. Do you want me to move the logic on 110 inside this for loop?

WillNilges: Oh wait I am foolish. NVM.

PR #384, opened 2024-06-02
Integrate django-import-export to allow bulk import and export

### LINKED ISSUE #201: Admins Need the Ability to "Bulk Create" install & member entries for large buildings
Slack discussion here: https://nycmesh.slack.com/archives/C06GW3Q92TY/p1708192421933169?thread_ts=1707797522.867909&cid=C06GW3Q92TY

Andrew-Dickinson:
Integrates a bulk import library to allow bulk import and export of objects. Our tables with custom primary key values need a slight customization, but nothing too bad

https://github.com/django-import-export/django-import-export

Tested export/import to/from CSV

Closes #201


WillNilges:
I thought that Django already had some kind of built in import/export tool. Do we really need to add another dependency?

https://docs.djangoproject.com/en/5.0/ref/django-admin/#dumpdata

https://docs.djangoproject.com/en/5.0/ref/django-admin/#loaddata

Andrew-Dickinson:
Those commands only allow you to export/replace large chunks of data like entire tables, and require CLI access and knowledge. This is a totally different use case?

Read the slack thread linked in the issue please

WillNilges:
Sorry, I thought that loaddata might be sufficient for this usecase, but if it's not, it's not. Let me take a look at the PR.

src/meshapi/admin/install.py
```diff
@@ -1,13 +1,27 @@
...
+from import_export import resources
+from import_export.admin import ExportActionMixin, ImportExportModelAdmin
 
 from meshapi.models import Install
 
 
+class InstallImportExportResource(resources.ModelResource):
+    def before_import(self, dataset: tablib.Dataset, **kwargs: Any) -> None:
+        if "install_number" not in dataset.headers:
+            dataset.headers.append("install_number")
```
WillNilges, line 17: Just out of curiosity, why this is necessary?

WillNilges (commented):
Seems reasonable. Can you add some docs on how to use it to the README?

Andrew-Dickinson:
I think the wiki is a better place for docs, since this is end-user focused not dev focused. But yeah I can write something up

src/meshapi/admin/install.py
Andrew-Dickinson: IIRC this is due to our custom primary key setup. Something about the lack of auto increment

Andrew-Dickinson:
https://wiki.mesh.nycmesh.net/books/6-services-software/page/how-to-bulk-create-installs-for-large-buildings

WillNilges (approved):

PR #385, opened 2024-06-02
Clean up add button on reverse inlines

Andrew-Dickinson:
Fix the spacing around the "Add" button that appears under Building inlines, and extend it to Install inlines also

### DIFF
--- a/src/meshapi/admin/inlines.py
+++ b/src/meshapi/admin/inlines.py
@@ -1,7 +1,7 @@
 from typing import Any, Optional
 
-import django.db.models
 from django.contrib import admin
+from django.contrib.admin import AdminSite
 from django.db.models import Model, Q, QuerySet
 from django.http import HttpRequest
 from nonrelated_inlines.admin import NonrelatedTabularInline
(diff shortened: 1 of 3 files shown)


src/meshapi/admin/inlines.py
```diff
@@ -135,3 +136,20 @@ class InstallInline(BetterInline):
...
+
+class InstallInlineBuildingPage(InstallInline):
+    add_button = True
+    reverse_relation = "building"
+    bulk_add_button = True
+
+
+class InstallInlineMemberPage(InstallInline):
+    add_button = True
+    reverse_relation = "member"
```
WillNilges, line 155: Do we really need to make a separate class for each of these? Can't we set the `reverse_relation` when we declare/construct it later?

Andrew-Dickinson: I don't think so. You can see how we use it below, we just pass a class reference in a list attribute on the `ModelAdmin` class. I can't think of a way to change this property that doesn't pollute the `InstallInline` model (other than the one I've implemented here)

WillNilges: Not even if we add a constructor?

WillNilges (approved):
Don't love having all these separate classes but we've got bigger fish to fry

src/meshapi/admin/inlines.py
Andrew-Dickinson: Oh no you're totally right. We're passed the model class in literally the first constructor param. Fixed

Issue #374, opened 2024-05-19
Join form doesn't do any phone number validation or standardization


Andrew-Dickinson:
Copy the standardization and validation logic from the import script to the join form

Andrew-Dickinson:
We should also add validation to API and admin UI edits

PR #388, opened 2024-06-20
Member phone validator

b-d055:
Attempt to address #374


Andrew-Dickinson:
Hey @b-d055. Thanks for your contribution. Sorry we didn't look at this earlier!

This change looks good to me, but we should have a few tests just to confirm it's behaving correctly. Could you add a unit test for:

1. Basic validation that creating a new model object and doing `model.save()` works with a valid phone number but fails with an invalid one
2. Submitting the join form with an invalid phone number throws an error
3. Attempting to change the phone number via the API fails with an invalid number

Andrew-Dickinson:
Also be sure to rebase onto main, we've made some linting changes so you'll want to double check that `invoke lint` is working as well

Andrew-Dickinson:
Closing this in favor of #450

PR #389, opened 2024-07-02
Add wireguard to deploy

james-otten:
Before this is merged, a list of variables need to be added.

If this is not done, I don't think deploys will work (given the config changes today).

### DIFF
--- a/.github/workflows/publish-and-deploy.yaml
+++ b/.github/workflows/publish-and-deploy.yaml
@@ -49,5 +49,15 @@ jobs:
         known_hosts: ${{ secrets.GRANDSVC_KNOWN_HOSTS }}
         #config: ${{ secrets.CONFIG }} # ssh_config; optional
         if_key_exists: fail # replace / ignore / fail; optional (defaults to fail)
+
+    - name: Setup WireGuard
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-and-deploy.yaml
```diff
@@ -49,5 +49,15 @@ jobs:
         known_hosts: ${{ secrets.GRANDSVC_KNOWN_HOSTS }}
         #config: ${{ secrets.CONFIG }} # ssh_config; optional
         if_key_exists: fail # replace / ignore / fail; optional (defaults to fail)
+
+    - name: Setup WireGuard
+      run:  |
+        run: sudo apt install wireguard
+        echo "${{ secrets.WIREGUARD_PRIVATE_KEY }}" > privatekey
```
Andrew-Dickinson, line 56: Is this really the correct way to do this? Won't it get echoed into the publicly readable logs?

james-otten: Invited you to a private repo where I tested something like this, I think it gets masked

Andrew-Dickinson (approved):

PR #396, opened 2024-07-06
Add "Autocomplete from ..." button to Device, Sector, Node and Building

### LINKED ISSUE #307: Auto populate lat/lon/alt for addresses entered by admins
The spreadsheet today automatically updates the lat/lon/alt columns based on Google geocoding when an address is changed. This is super helpful if you fix an address with a mistake in it. Maybe should auto populate the lat/lon/alt when admins create buildings manually?

We should also auto populate lat/lon/alt coordinates for admin-created devices, based on the node they are attached to

Andrew-Dickinson:
Adds a button to the following admin pages:

- Building
- Node
- Device
- Sector

When clicked, it automatically populates the geospatial coordinates (and DOB BIN where applicable) from the appropriate source. For Buildings, it uses the entered address (via a lookup to the city APIs). For Nodes, it uses the associated building (selecting the first one arbitrarily if there is more than one). For devices and sectors, it uses the associated node.

Example:

https://github.com/nycmeshnet/meshdb/assets/3138937/ff354541-3439-471f-9369-46887e604fcc

Closes #307


src/meshapi/static/widgets/auto_populate_location.js
```diff
@@ -0,0 +1,106 @@
...
+                            $('#id_altitude').val(absoluteAltitudeMeters.toFixed(1));
+                        })
+                    }
+
+                    data = {
+                        latitude: data.features[0].geometry.coordinates[1],
+                        longitude: data.features[0].geometry.coordinates[0],
+                        altitude: ""
+                    }
+                }
```
WillNilges, line 98: Can we re-use the logic from the join form here? Make an API call or something? I want to centralize and re-use it as much as possible, so that we don't end up having split-brain logic.

WillNilges (changes requested):

Andrew-Dickinson:
Screenshot 2024-08-04 at 18 27 24

src/meshapi/static/widgets/auto_populate_location.js
Andrew-Dickinson: Done!

src/meshapi/tests/test_geocode_api.py
WillNilges, line 1: A little odd that we're integration testing their API, but pop off.

src/meshapi/views/geography.py
```diff
@@ -365,3 +372,91 @@ def get(self, request: HttpRequest) -> HttpResponse:
...
+        if not serializer.is_valid():
+            return Response({"detail": serializer.errors}, status=status.HTTP_400_BAD_REQUEST)
+
+        try:
+            raw_addr: GeocodeRequest = serializer.save()
+            nyc_addr_info = geocode_nyc_address(raw_addr.street_address, raw_addr.city, raw_addr.state, raw_addr.zip)
+        except ValueError:
+            return Response(
+                {
+                    "detail": "Non-NYC registrations are not supported at this time. Check back later, "
```
WillNilges, line 439: Small Nit: Maybe we shouldn't say "check back later." I know we've been doing that for like a year, but it's probably more productive to just say "contact support."

WillNilges (approved):
Thank you!

src/meshapi/tests/test_geocode_api.py
Andrew-Dickinson: We're not integration testing their API. This is a mocked response where our testing framework pretends to be their API. It actually allows the opposite, for our tests to be fully independent of the city's API

src/meshapi/views/geography.py
Andrew-Dickinson: I can change this, but really this should be handled in the frontend and nobody should see this message

PR #397, opened 2024-07-06
Fix altitude parsing in join form logic

Andrew-Dickinson:
Altitude is reported by the city in feet, but we store it in meters, so we need to convert, and round to the nearest 0.1 meters when we do a lookup in the join form


WillNilges (approved):
Lol what an oversight

WillNilges:
Do you have a source you can link to about what units the city gives us the height in?

Andrew-Dickinson:
I don't have a source, I just spot checked a few addresses and they were totally off. Converting to meters made everything make sense

PR #401, opened 2024-07-07
Grant installers edit permissions on buildings & nodes

Andrew-Dickinson:
It seems odd to single out these two objects. I don't recall why we did that? Might just be a historical thing where we added these models and forgot to update the permissions?

This PR makes installer permissions consistent across all objects


WillNilges:
We did this because the Installer role was supposed to be for anyone with a baseline level of trust who would be setting up equipment on a rooftop. Right now, everyone who does that on a regular basis is pretty trustworthy, but I think the idea was that we didn't want that large of a group to mess with MeshDB information. We definitely didn't forget to update the permissions. 

I would prefer if we kept this role limited in scope and power.

WillNilges (commented):
I'd prefer if we didn't do this (See my comment)

Andrew-Dickinson:
Okay but like, they can trash the install, member, and device tables right? And they can't fix errors in a lat/lon position of a building or node. I think this is just going to result in unnecessary frustration. Like either we trust them to edit stuff or we don't, you know. I don't see singling out these two tables as something that offers all that much protection to us I guess

WillNilges:
Yeah, fair enough. I guess the installer role was some sort of premature optimization. We can always make more roles if we really feel strongly about it. I think restricting those tables might have been an attempt to prevent someone from potentially breaking info tied to many members (It's less bad if someone accidentally breaks a single Install vs a whole Node, you know?)

WillNilges (approved):

PR #410, opened 2024-07-14
Get pgadmin running securely™

WillNilges:
Just gonna try using traefik and an ACME cert instead of mess around with nginx


WillNilges:
https://wiki.mesh.nycmesh.net/books/6-services-software/page/pgadmin

james-otten (approved):

PR #414, opened 2024-07-20
Allow import of nn assigned statuses

Andrew-Dickinson:
I actually think due to some other logic, this might be dead code, but it is more logical and less misleading this way


WillNilges (approved):

WillNilges:
maybe add a comment?

PR #416, opened 2024-07-20
Add prod domain

james-otten:
_**db.mesh.nycmesh.net**_


Andrew-Dickinson:
Should we have `db.mesh` variants also?

WillNilges (approved):

WillNilges:
can we force db.mesh to redirect to https://db.mesh.nycmesh.net?

PR #417, opened 2024-07-22
[1-liner] Fix: crash on import due to incompatible types

Andrew-Dickinson:
Fix: type error causes crash for import of nodes without an install date


Andrew-Dickinson:
I'm bummed that `mypy` didn't catch this

WillNilges (approved):

PR #423, opened 2024-07-25
Add codecov run on merge to cover squashed commits

Andrew-Dickinson:
I think this is needed for the diffs to work correcty


Andrew-Dickinson:
That one line of code coverage diff is an artifact of some manual testing I did, and will not be an issue moving forward. I propose we override the failure and merge this anyway

WillNilges (approved):

WillNilges:
Sounds good to me. I wonder how annoying it might be to lose 0.03% on a given PR.

PR #427, opened 2024-07-28
Fix workflows

james-otten:
https://github.com/nycmeshnet/meshdb/actions/runs/10128636046

`The workflow is not valid. .github/workflows/deploy_prod1.yaml (Line: 10, Col: 5): The workflow must contain at least one job with no dependencies.`


WillNilges (approved):
We should figure out if we can avoid duplicating code like this, but I won't let perfect be the enemy of good :P

PR #430, opened 2024-07-29
Log scramble_members to stdout and add a y/n confirm prompt

WillNilges:
Need to explicitly set this so that users of the scramble command can actually see what it's doing.


Andrew-Dickinson (approved):

WillNilges:
I added a y/n confirmation prompt because I felt that may come in handy one day.

Andrew-Dickinson (approved):

PR #439, opened 2024-08-01
Redirect to prod domain

james-otten:
Redirect (302) to new k8s prod1

Docs: https://nginx.org/en/docs/http/ngx_http_rewrite_module.html


nginx/app.conf
```diff
@@ -5,12 +5,13 @@ server {
     access_log /var/log/nginx/access.log;
     error_log /var/log/nginx/error.log debug;
 
-    location = /favicon.ico { access_log off; log_not_found off; }
-    location /static/ {
-        root /var/www/html;
-    }
+    rewrite ^/(.*)$ https://db.mesh.nycmesh.net/$1 redirect;
```
Andrew-Dickinson, line 8: Does this work for `.mesh`?

Andrew-Dickinson (approved):

nginx/app.conf
james-otten: Does not take domain into consideration

Andrew-Dickinson:
Do we need to do extra work to redirect the other services? Do we need to redirect the other services?

james-otten:
> Do we need to do extra work to redirect the other services? Do we need to redirect the other services?

I wasn't planning on it.

Andrew-Dickinson:
> > Do we need to do extra work to redirect the other services? Do we need to redirect the other services?
> 
> I wasn't planning on it.

Feels like we should at least redirect meshforms since it’s user-facing. And at that point we should probably just redirect *.grandsvc? Redirects aren’t that hard right?

james-otten:
This is happening differently, Andrew is writing a config for apache on grand box to redirect all relevant services.

PR #441, opened 2024-08-01
Fix: import script crashes on datetime object

Andrew-Dickinson:
It's unclear why these keep surfacing (#392 being the other recent example). But for some reason we didn't catch these earlier? 

We definitely want date objects, not datetime ones, Django rejects datetime objects on date fields, idk why this ever worked


Andrew-Dickinson:
Also something mypy would probably catch if we didn't have it disabled on import stuff 😓

WillNilges (approved):

PR #443, opened 2024-08-01
Add django-cacheback to cache map JSON blobs for faster map render

Andrew-Dickinson:
This PR adds `django-cacheback` which allows the content of the map endpoints to be loaded asynchronously into redis for caching purposes. Then at request time, Django just pulls the JSON blob from Redis and serves it, saving expensive database calls

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -30,6 +30,8 @@ services:
       - api
     image:
       redis
+    ports:
+      - 6379:6379
 
   pgadmin:
     networks:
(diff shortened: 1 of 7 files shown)


Andrew-Dickinson:
Hmm I may abandon this effort. Even with the best case performance of the backend (cache hit, zero network overhead due to local hosting), we perform significantly worse than the existing map. I think looking into some kind of batch pre-computation solution similar to the existing map is probably the way to go...



https://github.com/user-attachments/assets/0ec75925-b5e8-48e7-bcf3-f6b67539883b

Andrew-Dickinson:
Confirmed with Willard on a call tonight, we're gonna kill this one and do this caching by pre-baking the map data in the npm build inside the docker container which serves the map, similar to the behavior of the original map

PR #444, opened 2024-08-01
Fix: NN assignment race condition for nodes created in admin UI

Andrew-Dickinson:
We were assigning NNs for nodes created in the admin UI without grabbing the lock which protects duplicate NN assignment. This fixes that :)


src/meshapi/models/node.py
```diff
@@ -90,10 +91,11 @@ class NodeType(models.TextChoices):
     )
 
     def save(self, *args: Any, **kwargs: Any) -> None:
-        if not self.network_number:
-            self.network_number = get_next_available_network_number()
+        with advisory_lock("nn_assignment_lock"):
```
WillNilges, line 94: So this will grab the lock any time you save _any_ node, even if you're not actually assigning an NN? Even if you're just saving notes?

That's probably fine, but perhaps consider moving the advisory lock to just cover the call to `get_next_available_network_number()`

Secondary Thought: Could we move that advisory lock call to `get_next_available_network_number` and save a few lines here and there?

Andrew-Dickinson: We cannot move the call inside of get_next_available_network_number() because then the race conditiion will still exist. The lock needs to cover from the moment we initiate the query to check the next available number, all the way until the number is used and saved to a database object (`super().save()` in this case)

We could maybe refactor to create a function which accepts a node object as an argument, mutates that, and saves it to the DB, all within the lock? 

That's also why the lock is so broad in this case, it reduces a `.save()` call, though I do see your point about potential performance impacts if the lock has to be held on every node save

Andrew-Dickinson: Pushed a change which reduce the lock scope a bit, lmk if you want further changes

WillNilges: Why will the race condition exist if it's within `get_next_available_network_number`? Can't we configure it to just block until whatever other process is finished?

WillNilges (approved):

WillNilges (approved):

src/meshapi/models/node.py
```diff
@@ -92,9 +92,10 @@ class NodeType(models.TextChoices):
 
     def save(self, *args: Any, **kwargs: Any) -> None:
         if not self.network_number:
-            with advisory_lock("nn_assignment_lock"):
-                self.network_number = get_next_available_network_number()
-                super().save(*args, **kwargs)
+            with transaction.atomic():
```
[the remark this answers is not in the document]
Andrew-Dickinson: We need both `transaction.atomic()` and `advisory_lock()` because they protect different things. `get_next_available_network_number()` uses `select_for_update()`, which requires the creation of a database transaction in order to signal release the selected Install object at the end of the transaction block. 

However, we also need to protect the in-memory copy of the lowest NN from races also, since `get_next_available_network_number()` can be called concurrently. It is true that `transaction.atomic()` protects a given code block from being entered more than once, but since we're calling `get_next_available_network_number()` from multiple places, that protection doesn't apply, so we use `advisory_lock()` for supplemental protection

PR #445, opened 2024-08-01
Fix: buildings can't be created in the admin UI

Andrew-Dickinson:
Buildings can't be created in the admin UI right now since it complains that `nodes` is a required field. You can't even set this field in the UI and it is definitely not required. This fixes that


WillNilges (approved):
Interesting. I guess the Join Form code is somehow able to overcome this.

Andrew-Dickinson:
Yeah the join form still works because it's not technically a django form object, so we bypass the Building form validation logic

PR #448, opened 2024-08-04
Install helm chart, reuse workflow, and remove node port

WillNilges:
See title. We're installing MeshDB with a helm chart now, as well as being a little more clever about re-using the workflow code. We're also removing the node port since we've got Traefik now.


james-otten (approved):

.github/workflows/publish-and-deploy.yaml
```diff
@@ -3,7 +3,7 @@ name: Publish and Deploy
 
 on:
   push:
-    branches: [ main ]
+    branches: [ wdn/reuse ]
```
james-otten, line 6: revert

.github/workflows/publish-and-deploy.yaml
```diff
@@ -63,111 +63,19 @@ jobs:
...
-                --set pgadmin.default_email=\"${{ secrets.PGADMIN_EMAIL }}\" \
-                --set pgadmin.default_password=\"${{ secrets.PGADMIN_PASSWORD }}\" \
-                --set uisp.user=\"${{ secrets.UISP_USER }}\" \
-                --set uisp.psk=\"${{ secrets.UISP_PSK }}\" \
-                | kubectl apply -f - && \
-              kubectl -n ${{ vars.APP_NAMESPACE }} rollout restart deploy \
-            '"
+    secrets: inherit
+    needs: push_to_registry
+    #if: github.ref == 'refs/heads/main'
```
james-otten, line 72: Lock back to main?

PR #449, opened 2024-08-04
Add endpoint to fetch LinkNYC kiosks

Andrew-Dickinson:
Adds an endpoint which replicates nodeDB's [kiosk fetch API](https://github.com/nycmeshnet/node-db/blob/master/src/kiosks.js) so that we can migrate off of it completely when we switch the website map to MeshDB


src/meshapi/views/map.py
```diff
@@ -16,6 +20,14 @@
     MapDataSectorSerializer,
 )
 
+LINKNYC_KIOSK_DATA_URL = "https://data.cityofnewyork.us/resource/s4kf-3yrf.json?$limit=100000"
```
WillNilges, line 23: Thought: Should we start an `apis.py` where we keep all of these API consts? That way they're in one place so there's no guessing about what we depend on externally.

src/meshapi/views/map.py
WillNilges, line 1: The plan is pretty much to murder this file when we write the new map, right?

WillNilges (approved):
Do we have a ticket tracking this, or was this something emergent with the old map? Doesn't matter that much to me, just curious.

pyproject.toml
```diff
@@ -19,6 +19,7 @@ dependencies = [
     "py3-validate-email@git+https://gitea.ksol.io/karolyi/py3-validate-email#egg=@v1.0.9",
     "phonenumbers==8.13.*",
     "requests==2.32.*",
+    "requests-mock==1.12.*",
```
james-otten, line 22: Optional dependency

src/meshapi/views/map.py
Andrew-Dickinson: I can do that, but as a separate PR. This one is big already

src/meshapi/views/map.py
Andrew-Dickinson: Yeah exactly. New map should be able to use the native REST APIs for the model objects, maybe we'll need a custom endpoint or two, but it will be for simple queryset filtering or something, none of this translation trash

pyproject.toml
Andrew-Dickinson: Good catch

src/meshapi/views/map.py
Andrew-Dickinson: https://github.com/nycmeshnet/meshdb/issues/453

Andrew-Dickinson:
> Do we have a ticket tracking this, or was this something emergent with the old map? Doesn't matter that much to me, just curious.

I noticed this as I was looking at [the map code](https://github.com/nycmeshnet/network-map/pull/66) based on the conversation we had over the weekend. Strictly speaking this isn't 100% necessary, since we could continue to depend on node-db for this one endpoint, but it seemed cleaner to pull everything under our new umbrella

PR #450, opened 2024-08-05
Phone number handling overhaul

Andrew-Dickinson:
Overhauls how we handle phone numbers, promoting them to first class identifiers of individuals. We now use them to de-duplicate at join form submission time, as well as at import script time.

We also now validate phone numbers at all ingress points, rejecting API calls that submit invalid numbers. We also normalize to a standard format, regardless of how the client provides the number, so that string comparisons are sufficient for uniqueness determination. 

Also overhauls join form de-duplication, including name change detection, which triggers a notification on slack for an admin to resolve


.env.sample
```diff
@@ -44,3 +44,5 @@ UISP_USER=nycmesh_readonly
 UISP_PASS=
 
 ADMIN_MAP_BASE_URL=http://admin-map.grandsvc.mesh.nycmesh.net
+
+SLACK_ADMIN_NOTIFICATIONS_WEBHOOK_URL=
```
WillNilges, line 48: I presume this isn't required

src/meshapi/tests/test_member.py
```diff
@@ -72,9 +72,83 @@ def test_member_all_phone_numbers_field(self):
...
+            code,
+            response.status_code,
+            f"status code incorrect. Should be {code}, but got {response.status_code}",
+        )
+
+        # Make sure the phone number is unchanged
+        test_member.refresh_from_db()
+        self.assertEqual(test_member.additional_phone_numbers, ["+1 456-555-6666"])
+
+        # Try to make a valid modification with a badly formated number and make sure it gets normalized
```
WillNilges, line 118:
```suggestion
        # Try to make a valid modification with a badly formatted number and make sure it gets normalized
```

src/meshapi/validation.py
```diff
@@ -29,15 +30,22 @@ def validate_email_address(email_address: str) -> bool:
     )
 
 
+def normalize_phone_number(phone_number: str) -> str:
+    return phonenumbers.format_number(
```
WillNilges, line 34: Were we.... not doing this before?

src/meshapi/validation.py
```diff
@@ -29,15 +30,22 @@ def validate_email_address(email_address: str) -> bool:
...
+        phonenumbers.parse(phone_number, "US"),
+        phonenumbers.PhoneNumberFormat.INTERNATIONAL,
+    )
+
+
 # Expects country code!!!!
```
WillNilges, line 40: Is this line true anymore? Can I just pass `555-555-5555` or `555 555 5555` and have it default to US?

src/meshapi/views/forms.py
```diff
@@ -117,24 +133,47 @@ def join_form(request: Request) -> Response:
             {"detail": "Your address could not be validated."}, status=status.HTTP_500_INTERNAL_SERVER_ERROR
         )
 
-    # Check if there's an existing member. Dedupe on email for now.
-    # A member can have multiple install requests
-    existing_members = Member.objects.filter(
-        Q(primary_email_address=r.email)
-        | Q(stripe_email_address=r.email)
-        | Q(additional_email_addresses__contains=[r.email])
+    # Check if there's an existing member. Group members by matching on both email and phone
```
WillNilges, line 136: Idea, could we gate creating new members in the admin panel with this logic?

src/meshapi/views/forms.py
```diff
@@ -92,13 +101,20 @@ def join_form(request: Request) -> Response:
             {"detail": "You must agree to the Network Commons License!"}, status=status.HTTP_400_BAD_REQUEST
         )
 
-    if not validate_email_address(r.email):
+    join_form_full_name = r.first_name + " " + r.last_name
```
WillNilges, line 104:
```suggestion
    join_form_full_name = f"{r.first_name} {r.last_name}"
```

src/meshapi/views/forms.py
```diff
@@ -117,24 +133,47 @@ def join_form(request: Request) -> Response:
...
+            phone_number=formatted_phone_number,
             slack_handle=None,
         )
     )
 
+    if r.email not in join_form_member.all_email_addresses:
+        join_form_member.additional_email_addresses.append(r.email)
+
+    if formatted_phone_number not in join_form_member.all_phone_numbers:
+        join_form_member.additional_phone_numbers.append(formatted_phone_number)
```
WillNilges, line 175: Just another brick in the data warehouse ✊🏻 😔 🧱

WillNilges: Actually tho do we want to treat this "new" contact info as more valuable? That is to say, replace the old primary with this one and slot the old primary into the `additional` field?

WillNilges: I think we've discussed this before and decided "no" because of griefing, but I wonder...

Maybe this is a problem for the "dashboard™". I could see a world where we send people texts/emails to verify them and only _then_ do we replace the primary.

WillNilges (commented):
A couple comments but mostly looks good.

Andrew-Dickinson:
Closes #374

src/meshapi/validation.py
WillNilges:
```suggestion
```

WillNilges: Make sure to add a test for interaction with different country codes.

src/meshapi/views/forms.py
WillNilges: Maybe also checking that data (email, phone, etc) you enter doesn't already belong to another Member.

src/meshapi/views/forms.py
WillNilges: maybe ask them to contact support if this happens?

src/meshapi/views/forms.py
Andrew-Dickinson: Added an issue to track this request https://github.com/nycmeshnet/meshdb/issues/463

src/meshapi/views/forms.py
Andrew-Dickinson: Discussed this on a call on Thursday, decided that this approach is fine for now, though it does expose a potential griefing opportunity of being able to add bogus email addresses to a member object. For now, this poses little security risk since these emails aren't really used for any kind of automated validation. 

We propose to securely validate identity using email/phone in the future if needed, we will require the use of the primary email address. You could imagine the logic being something like:

1. "Enter your email/phone to validate your account"
2. User enters non-primary email that _is_ in their additional_emails (which might be a bad email placed by an attacker) 
3. "oopsie, looks like that's not your primary email address, we have sent an email to p***********@gmail.com (primary address). Please click the link there to validate your identity"

This way we require the user to use the first email they ever used with us (or whatever an admin edits their primary email to) since this is very unlikely to have been hijacked

src/meshapi/validation.py
Andrew-Dickinson: Yes, we can pass plain US phone numbers now. Added tests to validate this (and that international phone numbers parse fine)

Andrew-Dickinson: Though I still think maybe we should require +1 on the join form? There's probably a nice little widget we can use to show a little country flag or something and do a bit of validation, etc

Andrew-Dickinson:
Closes #374

src/meshapi/util/admin_notifications.py
```diff
@@ -11,26 +12,40 @@
...
 SLACK_ADMIN_NOTIFICATIONS_WEBHOOK_URL = os.environ.get("SLACK_ADMIN_NOTIFICATIONS_WEBHOOK_URL")
+SITE_BASE_URL = os.environ.get("SITE_BASE_URL")
 
 
-def get_admin_url(model: Model) -> str:
+def get_admin_url(model: Model, site_base_url: str) -> str:
     content_type = ContentType.objects.get_for_model(model.__class__)
-    return reverse("admin:%s_%s_change" % (content_type.app_label, content_type.model), args=(model.pk,))
+    return urljoin(
+        site_base_url, reverse("admin:%s_%s_change" % (content_type.app_label, content_type.model), args=(model.pk,))
```
WillNilges, line 21: Can you f-string this?

src/meshapi/util/admin_notifications.py
```diff
@@ -11,26 +12,40 @@
...
     message: str,
-    request: HttpRequest,
+    request: Optional[HttpRequest] = None,
     raise_exception_on_failure: bool = False,
 ) -> None:
     serializer = serializer_class(model_instances, many=True)
 
+    site_base_url = SITE_BASE_URL
+    if request:
+        site_base_url = "{}://{}".format(request.scheme, request.get_host())
```
WillNilges, line 36: f-string?

WillNilges (approved):
f-strings

.env.sample
```diff
@@ -23,6 +23,8 @@ PELIAS_ADDRESS_PARSER_URL=http://localhost:6800/parser/parse
 NN_ASSIGN_PSK=localdev
 QUERY_PSK=localdev
 
+SITE_BASE_URL=https://db.mesh.nycmesh.net
```
Andrew-Dickinson, line 26: @james-otten for non-secret env variables is there a different procedure than the one from #455 ?

james-otten: https://github.com/nycmeshnet/meshdb/pull/476

Andrew-Dickinson:
Whoops I left auto-merge on 😓

PR #452, opened 2024-08-05
Add "Contact Preference" field to member object

had2020:
Requires members to indicate phone vs email contact preference via the join form, and store it on the member object.


src/meshapi_hooks/models.py
```diff
@@ -1,5 +1,31 @@
...
 
 # drf-hooks looks in this file for hook objects, but it feels weird to inline
 # it here, so we import it instead
 from .hooks import CelerySerializerHook
+class ContactPreference(models.TextChoices):
+    """Defines the available contact preference options."""
+    PHONE = 'phone', ('Phone')
+    EMAIL = 'email', ('Email')
+
+class Member(models.Model):
```
Andrew-Dickinson, line 12: I think you meant to modify the existing `Member` object which can be found [here](https://github.com/nycmeshnet/meshdb/blob/main/src/meshapi/models/member.py#L8)

src/meshapi_hooks/models.py
```diff
@@ -1,5 +1,31 @@
...
+    first_name = models.CharField(max_length=100)
+    last_name = models.CharField(max_length=100)
+    # ... other fields
+
+    contact_preference = models.CharField(
+        max_length=5,
+        choices=ContactPreference.choices,
+        blank=True,  # Allow null values for existing members
+        default=None,  # Avoid setting a default for existing members
+        help_text=("How would you like us to contact you?"),
```
Andrew-Dickinson, line 25: Creating a field like this is a good first step, but to make this actually useful we'll need to update the [join form handling logic](https://github.com/nycmeshnet/meshdb/blob/main/src/meshapi/views/forms.py#L82) to set this value, including changing the interface from to the frontend. We'll also need to update the [frontend code](https://github.com/nycmeshnet/meshforms/blob/main/components/JoinForm/JoinForm.tsx) to add an element to the form which allows the user to make a contact preference election, and pass their selection to the modified join form API endpoint

Andrew-Dickinson (changes requested):
I think you're on the right track, but we're definitely going to need a lot more changes before this is ready to merge. Let me know if you have any questions

had2020:
Thanks for the guidance, I will update the frontend coding for Django.

had2020:
I have finished my changes. I also added the element, in the frontend repo.

src/meshapi/views/forms.py
```diff
@@ -37,6 +37,7 @@ class JoinFormRequest:
     roof_access: bool
     referral: str
     ncl: bool
+    ContactPreference: str
```
Andrew-Dickinson, line 40: You will need to actually use this field in the logic below when constructing the member object, otherwise this will just get dropped on the floor

src/meshapi_hooks/models.py
```diff
@@ -1,5 +1,6 @@
 from django.db import models
+from django import forms
 
 # drf-hooks looks in this file for hook objects, but it feels weird to inline
 # it here, so we import it instead
-from .hooks import CelerySerializerHook
+from .hooks import CelerySerializerHook
```
Andrew-Dickinson, line 6: This probably won't pass the formatter. Please run `invoke lint` (and probably `invoke format`) before you submit to fix this sort of thing

src/meshapi/models/member.py
```diff
@@ -4,8 +4,21 @@
 from django.db.models.fields import EmailField
 from django_jsonform.models.fields import ArrayField as JSONFormArrayField
 
+class ContactPreference(models.TextChoices):
+    EMAIL = 'EMAIL', 'Email'
+    PHONE = 'PHONE', 'Phone'
+    SLACK = 'SLACK', 'Slack'
+    OTHER = 'OTHER', 'Other'
```
Andrew-Dickinson, line 11: I don't think we should provide Slack as an option here, since we don't currently have a way to collect slack handles (and this really doesn't make sense at join form time). I also think we should use null DB values rather than Other, since this more appropriately captures the unknown-ness. "Other" makes it sound like they have explicitly told us they don't want email or phone

src/meshapi/models/member.py
```diff
@@ -4,8 +4,21 @@
...
 
 class Member(models.Model):
+
+    contact_preference = models.CharField(
+        max_length=10,
+        choices=ContactPreference.choices,
+        default=ContactPreference.EMAIL,
+        help_text="The preferred method to contact this member"
```
Andrew-Dickinson, line 19: I would revise this to 

> The member's preferred contact method, likely indicated when they submitted the join form

src/meshapi/models/member.py
```diff
@@ -4,8 +4,21 @@
...
+class ContactPreference(models.TextChoices):
+    EMAIL = 'EMAIL', 'Email'
+    PHONE = 'PHONE', 'Phone'
+    SLACK = 'SLACK', 'Slack'
+    OTHER = 'OTHER', 'Other'
 
 class Member(models.Model):
+
+    contact_preference = models.CharField(
+        max_length=10,
```
Andrew-Dickinson, line 16: What does setting this parameter do for us? I don't think we're setting it anywhere else (for example `LinkType` on the `Link` model). Does this help with database size or something?

src/meshapi/models/member.py
```diff
@@ -4,8 +4,21 @@
...
+        choices=ContactPreference.choices,
+        default=ContactPreference.EMAIL,
```
Andrew-Dickinson, line 18: I think we should actually default to `None`, so that imported member objects don't get this automatically assigned

Andrew-Dickinson (changes requested):

PR #458, opened 2024-08-09
Fix: local dev doesn't work with Redis

Andrew-Dickinson:
Unit tests don't pass locally on main lol

Celery tries to connect to `redis://localhost:6379/0` but that port isn't forwarded to the system from the container


Andrew-Dickinson:
@james-otten you said I need to change something else when we modify this file?

james-otten:
> @james-otten you said I need to change something else when we modify this file?

Can we not merge this until after grand goes away.
I'm this case no chance needed.

Andrew-Dickinson:
Sure no worries

james-otten:
Already happened xD

Andrew-Dickinson:
Whoops :(

PR #465, opened 2024-08-11
Break celery worker and beat into their own pod

WillNilges:
Moves celery out of the main meshdb pod and into its own. The beat and worker each get their own container within that pod, and everyone communicates through Redis.

Inspired by: https://medium.com/ambient-innovation/health-checks-for-celery-in-kubernetes-cf3274a3e106


WillNilges:
https://medium.com/ambient-innovation/health-checks-for-celery-in-kubernetes-cf3274a3e106

Andrew-Dickinson (changes requested):
Can we move these celery related scripts into sub-directories? It's a lot of clutter for the root of the repo

Andrew-Dickinson:
Happy to override codecov here. Testing celery statically is hard. With Datadog will we be able to monitor to ensure the tasks are running and alert if they stop?

.github/workflows/publish-and-deploy.yaml
```diff
@@ -3,7 +3,7 @@ name: Publish and Deploy
 
 on:
   push:
-    branches: [ main ]
```
Andrew-Dickinson, line 6: Did you mean to leave this changed?

.github/workflows/publish-and-deploy.yaml
```diff
@@ -69,7 +69,7 @@ jobs:
       environment: dev3
     secrets: inherit
     needs: push_to_registry
-    if: github.ref == 'refs/heads/main'
```
Andrew-Dickinson, line 72: This too?

entrypoint.sh
```diff
@@ -11,19 +11,6 @@ done
 
 echo 'DB started'
 
-# It's okay to start Celery in the background and continue without waiting, even though "migrate"
-# might make DB changes we want to notify for since tasks are queued by Django Webhook and
-# are executed as soon as celery starts
-# FIXME: This makes testing locally a bit awkward, since this isn't started by "manage.py runserver"
-#  maybe there's a way to do this better?
```
Andrew-Dickinson, line 18: How does this comment change with these changes? Is the guidance still to run these commands from the readme for local dev?
```
celery -A meshdb worker -l INFO
celery -A meshdb beat -s /tmp/celerybeat-schedule -l DEBUG
```

Also does the celery worker we spin up programmatically during unit testing need to be changed in any way?

WillNilges: Shouldn't be. This just moves the worker and beat out of the prod container. Grandbox will change, but whatever. I guess what I'd do if I were testing this locally would be to run the provided scripts.

WillNilges:
> Can we move these celery related scripts into sub-directories? It's a lot of clutter for the root of the repo
> 
> Perhaps:
> 
> * bin/
>   
>   * celery/
>     
>     * All the celery stuff
>   * All the other scripts we're collecting

@Andrew-Dickinson I agree. I was going to make a `scripts/` directory in a follow-up PR, but I'm happy to do it in this one.

.github/workflows/publish-and-deploy.yaml
WillNilges: Gosh you're quick on the draw

Andrew-Dickinson: Fastest click in the west 🤠

WillNilges:
> Happy to override codecov here. Testing celery statically is hard. With Datadog will we be able to monitor to ensure the tasks are running and alert if they stop? (and can we document that requirement, in #88 perhaps)

Certainly. Feel free to add that to the reqs.

entrypoint.sh
Andrew-Dickinson: Could you update the readme with the latest guidance?

WillNilges: Fuck, I knew you were gonna say that.

WillNilges: Actually, on second thought, the readme instructions are fine. What's there right now will still work.

src/meshdb/celery.py
```diff
@@ -13,37 +13,37 @@
 class LivenessProbe(bootsteps.StartStopStep):
     requires = {"celery.worker.components:Timer"}
 
-    def __init__(self, worker, **kwargs):
+    def __init__(self, worker, **kwargs):  # type: ignore[no-untyped-def]
```
Andrew-Dickinson, line 16: We can probably do types here, just use `dict` for the `kwargs` and `None` for the return type

Andrew-Dickinson (approved):
Approved, but I do think you can do better than that on the MyPy types and I would like to see _some_ improvement over what we have now.

If you're really going to ignore that many lines you should probably just exclude the whole file

src/meshdb/celery.py
WillNilges: Didn't know that, thanks.

Andrew-Dickinson: Worker may be tricky but if it has a class you can just use the class and it’s essy

WillNilges:
_maybe_ we can get the redefinitions to stop if we break out the checks into other files, like in the blog post I attached in the description.

PR #466, opened 2024-08-12
Add Celery Worker to Automatically Import from UISP

### LINKED ISSUE #383: Figure out what ongoing UISP import looks like
How are we going to handle ongoing synchronization between UISP and meshDB?

What happens when a device name is changed in UISP to one value and another in meshDB?

Andrew-Dickinson:
We should now be able to synchronize automatically with UISP on an hourly basis

When there is a diff with MeshDB, we take the UISP values, and we notify admins of the fact this has occurred, via the new slack notification system introduced in #450

Stole some import logic from the existing import script. However, a lot of that is tightly coupled to spreadsheet nonsense and should be shipped out to sea, which means most of this is new code anyway.

Also adds types for the UISP API, made in collaboration with ChatGPT :) 

Closes #383


Andrew-Dickinson:
Hmm perhaps the types should be excluded from coverage information

src/meshapi/util/uisp_import/sync_handlers.py
```diff
@@ -0,0 +1,272 @@
...
+                f"because the inferred NN ({uisp_network_number}) did not correspond to any "
+                f"nodes in our database"
+            )
+            continue
+
+        if not uisp_device["overview"]["status"]:
+            # If UISP doesn't have a status value, assume active
+            uisp_status = Device.DeviceStatus.ACTIVE
+        elif uisp_device["overview"]["status"] == "active":
+            uisp_status = Device.DeviceStatus.ACTIVE
```
WillNilges, line 76: Nit: Could you do this with an `or` in the above clause?

src/meshapi/models/los.py
```diff
@@ -38,6 +38,9 @@ class LOSSource(models.TextChoices):
     )
 
     analysis_date = models.DateField(
+        default=None,
+        blank=True,
+        null=True,
```
WillNilges, line 43: I know this has something to do with the unreliable data we have about our existing links, but why would we want this to be nullable? What was the impetus for this decision?

WillNilges (commented):
First round of comments. Full disclosure: I'm tired.

src/meshapi/types/uisp_api/data_links.py
```diff
@@ -0,0 +1,144 @@
+# Auto-generated by ChatGPT from the UISP API Swagger file available here
+# https://uisp.mesh.nycmesh.net/nms/api-docs/swagger.json
+# with small manual tweaks to correct its mistakes. If these are wrong don't @ me
```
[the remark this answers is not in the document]
Andrew-Dickinson: The correctness of these types is impossible to automatically test without _extensive_ work. The UISP devs themselves may not even be able to do it easily

That being said, the consequence for having wrong types here is extremely minor. You'll get an annoying warning from mypy about a property not existing (which will be contradicted by the UISP docs). But this issue is actually guaranteed to be the case at some point for at least some properties (since UISP is likely to add new fields to their API in the future). The way to get unblocked would just be to add the missing type (as documented by Ubiquiti) and the mypy error will go away

These types aren't even used for runtime validation, they're really just a connivence feature for devs (and a way to make mypy complain less). If you really don't want them in here, my fallback plan would be to just mark everything as `dict` and give up on type definitions for UISP responses. On the whole I think keeping the ChatGPT types leans into more benefit than harm, but I do see where you are coming from in not trusting the AI farther than you can unit test it.

I marked this file as excluded from the codecov reports because it was juicing our stats, but FWIW all the lines were green (though this just indicates they are imported _somewhere_, not that useful)

src/meshapi/tasks.py
```diff
@@ -33,13 +39,30 @@ def run_update_panoramas() -> None:
...
         "schedule": crontab(minute="0", hour="*/1"),
     },
+    "import-from-uisp-hourly": {
+        "task": "meshapi.tasks.run_update_panoramas",
+        "schedule": crontab(minute="10", hour="*/1"),
+    },
+    "run-database-backup-hourly": {
+        "task": "meshapi.tasks.run_database_backup",
+        "schedule": crontab(minute="20", hour="*/1"),
+    },
```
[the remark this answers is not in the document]
Andrew-Dickinson: Good question. I don't know, so probably not. Adding a task to try to profile these UISP changes at the very least

Tell me more about the file touch thing, I'm not familiar

src/meshapi/models/los.py
Andrew-Dickinson: This boils down to basically the same discussion we had about whether or not to make email address nullable since it's so core to our processes (can't find the issue, maybe it was on a call?) 

The reality is that for some of these LOS entries, there is a small chance we just don't have a date on the upstream Link object that they have been generated from. We can make up a bullshit date in these cases (1970-01-01 or the founding date of the mesh or something), but to me it feels better to pass the `null` along to the consumer so that they can make an educated choice (and not do silly things with our silly date). Fundamentally if we don't have the data we don't have the data

Within meshdb this field isn't currently used for anything, we just carry it along as metadata. Likely future use cases would be things like filtering out old LOS analyses when trying to asses LOS impacts of a new building or something

Perhaps we should actually just do `null=True`, without setting `default` or `blank`. This would force admins who try to create/modify LOS objects with admin forms to always provide a date. Though this includes cases where they modify an auto-generated object that did not have a date before they touched it. That might be annoying on the one hand, but rare, and on the other hand they can probably make a higher quality guess than the automated system can

Andrew-Dickinson: Discussed on a call: we're going to keep it not required in the admin forms, but on the create form, auto fill the current date

src/meshapi/tasks.py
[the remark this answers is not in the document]
Andrew-Dickinson: Discussed on a call. Since there is not precise sequencing (or mutual exclusivity) required between these workers, we should be fine to leave this as it is, indicating a light preference for them to happen at different times (and for backups to happen last), but not enforcing it

I will verify that the data accesses in the UISP sync use `SELECT FOR UPDATE` in the right places, so that we don't have race conditions (against these other workers or against API/admin traffic)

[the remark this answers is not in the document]
Andrew-Dickinson: Patched a potential race condition, should be safe to run concurrently with anything now

src/meshapi/util/uisp_import/fetch_uisp.py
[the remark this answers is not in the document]
Andrew-Dickinson: Yeah it isn't covered because the tests would be really silly, just like checking that each function makes an http call, but coverlay is angry that these don't get covered by any tests (since these are used as the mocking point for the import stuff)

Andrew-Dickinson:
Ran the celery task locally. Seems to work, and takes ~5 minutes on my laptop

src/meshapi/types/uisp_api/devices.py
WillNilges, line 1: If I was mean I would suggest breaking all these types out into a separate library or something. I really can't believe that UISP doesn't have any sort of lib for this.

src/meshapi/types/uisp_api/literals.py
WillNilges, line 1: Jesus, this is all Swagger AI vomit?

src/meshapi/util/uisp_import/sync_handlers.py
```diff
@@ -0,0 +1,297 @@
...
+                f'(UISP ID {uisp_uuid}) was skipped because the "{uisp_category}" has been '
+                "excluded in meshapi.util.uisp_import.constants"
+            )
+            continue
+
+        modified_name = uisp_device["identification"]["name"]
+        for find, replace in DEVICE_NAME_NETWORK_NUMBER_SUBSTITUTIONS.items():
+            modified_name = modified_name.replace(find, replace)
+
+        network_number_matches = re.findall(NETWORK_NUMBER_REGEX_FOR_DEVICE_NAME, modified_name)
```
WillNilges, line 53: Oh boy, a regex. I see that you've got tests for this function, so I guess I'll let it slide. Can you show me some examples of what the regex is supposed to match?

src/meshapi/util/uisp_import/sync_handlers.py
```diff
@@ -0,0 +1,297 @@
...
+            notes=f"Automatically imported from UISP on {datetime.date.today().isoformat()}\n\n",
+            # UISP doesn't track the following info
+            install_date=None,
+            abandon_date=None,
+            description=None,
+        )
+        link.save()
+
+
+def sync_link_table_into_los_objects() -> None:
```
WillNilges, line 253: Oh man it kinda sucks that we have to do this. I wonder if a Link could have a mandatory LOS field? Maybe that wouldn't work for things like fiber...

WillNilges (approved):
yea seems reasonable. Left a couple of comments.

src/meshapi/types/uisp_api/devices.py
Andrew-Dickinson: Separate library makes me want to cry :(

How about this? I'll break it out into a separate library when I (inevitably) need these types in another project

src/meshapi/types/uisp_api/literals.py
Andrew-Dickinson: Affirmative. Though the possible values are just copy-pasted from the swagger file. All the AI really did was reformat things

src/meshapi/util/uisp_import/sync_handlers.py
Andrew-Dickinson: It extracts the the NN from the device name, which is the only way I can think of to do the mapping. This is a bit tricky for p2p radios, but fortunately the convention is that the NN the device is located at is the first number, and the second number is the remote side. Some examples:
```
nycmesh-227-north -> 227
nycmesh-5916-lbe-3606 -> 5916
nycmesh-lbe-238 -> 238
``

File: src/meshapi/util/uisp_import/sync_handlers.py (hunk 6 above)
Comment (Andrew-Dickinson) [reply src/meshapi/util/uisp_import/sync_handlers.py:253]: Strictly speaking we don't have to do this, it's just going to make the LOS table more useful to do it. For the map to render accurately we don't need any LOS objects that duplicate links. 

Regarding some kind of foreign key with link, I think that could possibly be useful, but I think it would have to be m2m and optional (fiber is a great example of this as you point out). It feels like for our current needs it is overkill and will be more trouble than it's worth, but perhaps that will change as we work towards Map v2 TM (and probably LOS features under the same umbrella).

Good call about fiber though! We should definitely exclude fiber and ethernet links from our assumption that "every link implies a LOS"

Comment (Andrew-Dickinson) [reply src/meshapi/util/uisp_import/sync_handlers.py:253]: Also VPN 🫣

File: src/meshapi/util/uisp_import/sync_handlers.py [hunk 7]
```diff
@@ -0,0 +1,301 @@
...
+                f"which we do not have in our database (perhaps it was skipped at device import time?)"
+            )
+            continue
+
+        if uisp_link["state"] == "active":
+            uisp_status = Link.LinkStatus.ACTIVE
+        else:
+            uisp_status = Link.LinkStatus.INACTIVE
+
+        uisp_link_type = get_link_type(uisp_link)
```
Andrew-Dickinson, line 213: TODO: This type can be incorrect when links get added manually in the UISP web console. Notably this results in fiber and VPN links loading as "ethernet"

To make things worse, if this is a new link, we won't even tell the admins about it so they can fix it....

WillNilges (approved):

PR #474, opened 2024-08-15
Add buttons to view devices in UISP and Installs in OSTicket

### LINKED ISSUE #404: Link to UISP from the device page
Create a button similar to the IP address link on the UISP ID field that takes you to the UISP page for that device

Maybe do links also?

### LINKED ISSUE #419: Link to OSTicket from the admin page
The install admin page should have a link to the OSTicket entry, similar to how the IP address from #289 have a little eye icon. Should be similar to #403 and #404 as well

WillNilges:
What it says on the tin, plus we now have an ExternalHyperlinkWidget that we can use to create more of these for other services if it comes up.

Closes #419
Closes #404


src/meshapi/admin/device.py
```diff
@@ -6,15 +8,21 @@
 
 from meshapi.admin.inlines import DeviceLinkInline
 from meshapi.models import Device
-from meshapi.widgets import UISPHyperlinkWidget
+from meshapi.widgets import ExternalHyperlinkWidget
+
+UISP_URL = os.environ.get("UISP_URL", "https://uisp.mesh.nycmesh.net")
```
Andrew-Dickinson, line 13: The current value in `.env.sample` (and expected by the import script) is

```
UISP_URL=https://uisp.mesh.nycmesh.net/nms
```

so you'll have `/nms` in the url twice with this setup

src/meshapi/widgets.py
```diff
@@ -24,6 +24,7 @@ def pano_get_context(self, name: str, value: str) -> dict:
             }
         }
 
+    # FIXME (Use SafeString)
```
Andrew-Dickinson, line 27: todo?

src/meshapi/templates/widgets/external_link.html
```diff
@@ -0,0 +1,16 @@
+{% load i18n static %}
+<div class="related-widget-wrapper">
+    {% include "django/forms/widgets/input.html" %}
+    {% block links %}
+    {% spaceless %}
+    <a class="related-widget-wrapper-link view-related"
+        {% if widget.value %}
+            href="{{ widget.formatted }}"
+        {% endif %}
```
Andrew-Dickinson, line 9: Ux nit: make this if block control a CSS style instead, so the icon is like 50% opacity instead of missing

src/meshapi/admin/install.py
```diff
@@ -9,6 +10,9 @@
 from import_export.admin import ExportActionMixin, ImportExportModelAdmin
 
 from meshapi.models import Install
+from meshapi.widgets import ExternalHyperlinkWidget
+
+OSTICKET_URL = os.environ.get("OSTICKET_URL", "https://support.nycmesh.net")
```
Andrew-Dickinson, line 15: Please add this to `.env.sample` (and the various IAC bits. @james-otten made an example here #476)

Andrew-Dickinson (changes requested):
Love the reusable widget

src/meshapi/admin/device.py
WillNilges: I think we should change it tbh.

src/meshapi/templates/widgets/external_link.html
WillNilges: Not sure I understand. This is what it looks like when the value is missing: 
image

src/meshapi/admin/install.py
WillNilges: It looks like this is already added in prod.

src/meshapi/templates/widgets/external_link.html
Andrew-Dickinson: Oh nvm looks great

Andrew-Dickinson:
Can you add this to links also? they also have a UISP ID

Does what you've done appear on sectors and APs also?

WillNilges:
> Can you add this to links also? they also have a UISP ID
> 
> Does what you've done appear on sectors and APs also?

Can you give me an example of what a link looks like in UISP? Pulling up this link isn't working for me: https://db.mesh.nycmesh.net/admin/meshapi/link/609/change/?_changelist_filters=q%3D365

Andrew-Dickinson:
Oh bummer you’re right I don’t think it supports deep-linking to links :(

Andrew-Dickinson:
The best we could do would be to call the UISP API and describe the link and then hyperlink to one of the endpoints, but gross

WillNilges:
I can make it work on sectors. I can also add it to access points, but I don't see any UISP IDs for them, at least in my copy of the DB.

Andrew-Dickinson (approved):

PR #478, opened 2024-08-19
Fix: crash on import due to foreign key issue

Andrew-Dickinson:
Import was crashing because the trick I used to change the ID of an object doesn't work when other objects reference it, kinda makes sense. Instead I use a trick where we construct a new object with the same ID as the one we want to replace, and then just call `.save()` to force the update. Easy


james-otten (approved):
_Tricky_

PR #483, opened 2024-08-23
Set urls from env + cleanup

james-otten:
Set user facing environment-specific urls based on the `MESHDB_ENVIRONMENT` env var. Also, make Willard more happy.

### DIFF
--- a/src/meshdb/settings.py
+++ b/src/meshdb/settings.py
@@ -36,25 +36,10 @@ PROFILING_ENABLED = DEBUG and not os.environ.get("DISABLE_PROFILING", "False") =
 
 USE_X_FORWARDED_HOST = True
 
-MESHDB_ENVIRONMENT = os.environ.get("MESHDB_ENVIRONMENT", "")
-if MESHDB_ENVIRONMENT == "":
-    # Local or grand
-    LOS_URL = "https://los.grandsvc.mesh.nycmesh.net"
-    MAP_URL = "https://map.grandsvc.mesh.nycmesh.net"
-    PG_ADMIN_URL = "/pgadmin/"
-    FORMS_URL = "https://forms.grandsvc.mesh.nycmesh.net"
-elif MESHDB_ENVIRONMENT == "prod":
-    # k8s prod
-    LOS_URL = "https://los.mesh.nycmesh.net"
-    MAP_URL = "https://map.mesh.nycmesh.net"
-    PG_ADMIN_URL = "/pgadmin/"
-    FORMS_URL = "https://forms.mesh.nycmesh.net"
-elif MESHDB_ENVIRONMENT == "dev3":
-    # k8s dev3
-    LOS_URL = "https://devlos.mesh.nycmesh.net"
-    MAP_URL = "https://devmap.mesh.nycmesh.net"
-    PG_ADMIN_URL = "/pgadmin/"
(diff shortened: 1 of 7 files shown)


src/meshdb/settings.py
```diff
@@ -36,6 +36,26 @@
 
 USE_X_FORWARDED_HOST = True
 
+MESHDB_ENVIRONMENT = os.environ.get("MESHDB_ENVIRONMENT", "")
```
WillNilges, line 39: Interesting idea. How will this work for local dev?

src/meshweb/views.py
```diff
@@ -13,15 +14,15 @@ def index(request: HttpRequest) -> HttpResponse:
     template = loader.get_template("meshweb/index.html")
     links = {
         "Member Tools": [
-            ("https://forms.grandsvc.mesh.nycmesh.net/join/", "Join Form"),
-            ("https://los.grandsvc.mesh.nycmesh.net/", "Line of Sight Tool"),
-            ("https://map.grandsvc.mesh.nycmesh.net", "Map"),
+            (f"{settings.FORMS_URL}/join/", "Join Form"),
```
WillNilges, line 17: omg thank you for doing this

src/meshdb/settings.py
james-otten: Defaults to same as before for grand and local

src/meshweb/views.py
james-otten: :)

Andrew-Dickinson:
I want to look at this please. Cool idea but I don’t want to loose functionality

src/meshdb/settings.py
```diff
@@ -36,6 +36,26 @@
...
+if MESHDB_ENVIRONMENT == "":
+    # Local or grand
+    LOS_URL = "https://los.grandsvc.mesh.nycmesh.net"
+    MAP_URL = "https://map.grandsvc.mesh.nycmesh.net"
+    PG_ADMIN_URL = "/pgadmin/"
+    FORMS_URL = "https://forms.grandsvc.mesh.nycmesh.net"
```
Andrew-Dickinson, line 45: Hmm there is overlap here in functionality with several values in `.env`:
- `SITE_BASE_URL`
- `ADMIN_MAP_BASE_URL`
- `PELIAS_ADDRESS_PARSER_URL`
- `UISP_URL`

Should we consolidate these other URLs to this style? I'm worried about having two different systems and it not being clear which one should be used for a given URL. It seems like you only changed this for the links shown to users (and the ones from `.env` are values used internally). I think having that as the standard is fine but I do wish there was a way to make it a bit more clear for newer folks.

Adding these `.env` URLs here would also introduce the issue of local testing overrides causing a `git diff` (a pet peeve of mine). Perhaps they could be included here and the `.env` values could act as an override?

Andrew-Dickinson: Also perhaps this should default to dev rather than grandbox, since I thought we were putting grandbox out to pasture

james-otten: Done, please have a look

james-otten: So many clicks

Andrew-Dickinson (approved):
Cool. For the record I’m not opposed to this “one env variable controls many settings constants” pattern so if @WillNilges wants it we can do it. Mostly I just wanted an escape hatch for local dev

Issue #217, opened 2024-02-24
Investigate Admin Panel tests


WillNilges:
We need something that will let us test the AdminUI itself. #215 fixes a bug that is difficult (if not impossible) to catch in our current testing setup because it has to do with the templates. There was a problem with one of the templates causing reverse FK fields to be blanked out on the building and member `change` routes.

https://www.selenium.dev/selenium-ide/ might be a decent candidate.

Andrew-Dickinson:
Selenium is powerful but very heavy (bundles chromium). We only need it if we want to run the javascript, which on the admin page doesn't do very much. For just asserting things about the generated HTML responses, using [beautifulsoup](https://beautiful-soup-4.readthedocs.io/en/latest/) should be much easer to do

WillNilges:
we could use datadog for this <3

PR #492, opened 2024-08-26
Fix: node form is not submittable and add admin tests

### LINKED ISSUE #490: Node objects can't be edited in the admin UI
Screenshot 2024-08-24 at 00 43 28

No further info is provided, probably we need to set a 'default' or a 'blank' somewhere

Andrew-Dickinson:
Fixes #490 

Also adds missing import/export options to LOS in the Admin UI

And adds more comprehensive testing for search, list, view, and change in the admin UI


Andrew-Dickinson:
Fixes: #217. Depending on your perspective, @WillNilges. These tests now cover all basic interactions with the admin UI. Perhaps we should add some more tests to cover the detailed behavior of the inlines and that is what we would want to see to close #217? I would argue that's a post-launch nice-to-have

WillNilges (approved):

PR #497, opened 2024-08-30
Dramatically improve search relevancy using Postgres' search ranking features

### LINKED ISSUE #470: Improve building search
Searching by addresses, i.e. `91 E 3rd St` returns very poor relevancy results

### LINKED ISSUE #477: Rank exact matches higher than partial matches in the admin search
When trying to find NNs like `3` or `10`, it is very difficult because the nn you are looking for is drowned in irrelevant results. We should rank exact matches at the top to avoid this

I think this principle applies in general to other things, like install numbers, and maybe zip codes? Completion of this issue should include an analysis of other searches that could benefit from this kind of improvement

Andrew-Dickinson:
Use the `SearchRanking` class to rank search result relevancy in the admin UI. This makes an incredible difference. See examples in thread below

Also made some small tweaks to the fields included in the search, to remove things that were harming relevancy without adding much value

Pro tip: this PR also includes a refactor, so the diff is a bit messy. Step through commit-by-commit and it will make more sense

Closes #477 
Closes #470


Andrew-Dickinson:
|        Before            |       After         |
:-------------------------:|:-------------------------:
| Screenshot 2024-08-29 at 22 28 36 | Screenshot 2024-08-30 at 01 12 18
| Screenshot 2024-08-29 at 22 28 49 | Screenshot 2024-08-30 at 01 12 02
Screenshot 2024-08-29 at 22 29 00 | Screenshot 2024-08-30 at 01 12 27
Screenshot 2024-08-30 at 01 23 02 | Screenshot 2024-08-30 at 01 22 09
Screenshot 2024-08-30 at 01 27 29 | Screenshot 2024-08-30 at 01 27 51

src/meshapi/admin/__init__.py
WillNilges, line 1: Thanks, but why do this? Consistency with other modules?

src/meshapi/admin/models/access_point.py
```diff
@@ -30,9 +32,10 @@ class Meta:
 
 
 @admin.register(AccessPoint)
-class AccessPointAdmin(ImportExportModelAdmin, ExportActionMixin):
+class AccessPointAdmin(RankedSearchMixin, ImportExportModelAdmin, ExportActionMixin):
     form = AccessPointAdminForm
-    search_fields = ["name__icontains", "notes__icontains"]
+    search_fields = ["name__icontains", "@notes"]
+    search_vector = SearchVector("name", weight="A") + SearchVector("notes", weight="D")
```
WillNilges, line 38: Is there a difference between doing `A, D` and `A, B` if there's only two options?

src/meshapi/admin/models/building.py
```diff
@@ -56,29 +59,31 @@ class Meta:
...
         "nodes__network_number__iexact",
         "installs__install_number__iexact",
-        # Search by Member info
-        "installs__member__name__icontains",
-        "installs__member__primary_email_address__icontains",
-        "installs__member__phone_number__iexact",
-        "installs__member__slack_handle__iexact",
         # Notes
-        "notes__icontains",
+        "@notes",
```
WillNilges, line 76: What does this mean vs `notes__icontains`?

src/meshapi/admin/ranked_search.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import Optional, Tuple
+
+from django.contrib.admin import ModelAdmin
+from django.contrib.admin.views.main import ORDER_VAR
+from django.contrib.postgres.search import SearchQuery, SearchRank, SearchVector
+from django.db.models import QuerySet
+from django.db.models.expressions import CombinedExpression
+from django.http import HttpRequest
+
+# Trick stolen from https://stackoverflow.com/a/56991089 to make mypy happy about mixin types
```
WillNilges, line 11: I wonder if I could use that on the button thing I was having trouble with the other day.

WillNilges (approved):

src/meshapi/admin/__init__.py
WillNilges: After reviewing the PR, I think I understand.

Andrew-Dickinson: I initally refactored to try to solve a circular import issue I was seeing, but ultimately I kept it because it felt more organized

src/meshapi/admin/ranked_search.py
Andrew-Dickinson: I don't think so. I thought about this then. The reason this is needed is that for MixIns, the parent class is unknown to the mypy at declaration time, so we use this trick to tell it what the parent class will eventually be. In your situation, mypy knew the parent class, but was wrong about the availability of its methods

src/meshapi/admin/models/access_point.py
Andrew-Dickinson: Yes, from the [Django docs on this feature](https://docs.djangoproject.com/en/5.1/ref/contrib/postgres/search/#postgresql-fts-weighting-queries):

> The weight should be one of the following letters: D, C, B, A. By default, these weights refer to the numbers 0.1, 0.2, 0.4, and 1.0, respectively

So by doing `A, D` instead of `A, B`, we weight the notes field about 4x lower. This gives a "fuzzy match" on the name field, which otherwise would get a low score an even better chance of winning against an exact match in the notes of a separate object. Effectively it weights the name *even higher* than the notes

I'm not 100% sure this is *always* the best. I can't find the example I worked off of now, but it seems to be the way people do this sort of "headine" vs "body contents" thing in other applications

src/meshapi/admin/models/building.py
Andrew-Dickinson: The `@` symbol instructs Django to construct the queryset with `__search`, which uses [postgres' full-text search features](https://docs.djangoproject.com/en/5.1/ref/contrib/postgres/search/#the-search-lookup). Some of this is [over my head](https://www.postgresql.org/docs/current/datatype-textsearch.html#DATATYPE-TSVECTOR), but I think basically it does more fuzzy matching, and does some search-engine style splitting of the input so that the precise placement of punctuation, etc. doesn't totally blow up your results. As I understand it, `__icontains` is good for short strings where you're likely to want exact matches, but `__search` is better for long blocks of text where the user is unlikely to remember the exact details character-for-character

Issue #489, opened 2024-08-24
Potential Hubs and Supernodes are not shown in MeshDB's map


Andrew-Dickinson:
So whoops, in a similar vein to #402 we are unable to represent potential hubs and supernodes without assigning an NN. This is obviously bad, since we can't show the correct potential dots on the map for these sites

Andrew-Dickinson:
Options:
1 ) Go back to parsing the word `hub` that we imported from the spreadsheet from the install notes in order to tell the map that this install is a potential hub. Upsides: easy and fast. Downsides: brittle and unpredictable
2) Introduce a new field on install which allows us to select "node type" similar to how we do this with Nodes. Upsides: unambiguous. Downsides: duplicative of the field on node, going to be unused for active nodes
3) Assign NNs for potential hubs? Upsides: simple, few code changes. Downsides: inefficient use of NN allocations
4) Change the primary key for the node object to something other than NN, allowing potential Node objects without assigning an NN. Upsides: no data duplication. Downsides: APIs and URLs get very confusing
4a) Assign potential hubs super high "fake" NNs like `100000xxxx`

Andrew-Dickinson:
Perhaps we could do option 4, but keep all existing URL and API interfaces the same (this might be hella painful in the admin UI though IDK). Basically the only place the real `pk` would be used would be internal to the db, but the URLs would all reference the network number

WillNilges:
After a phonecall, we had two thoughts:
- Add two fields to the Install: "Planned Hub" and "Planned Supernode" which will control the size and color of the dot
- Add a union field between name and NN, where the name field is for when you don't have an NN, but if you assign it, it gets clobbered.

But, the better thing to do would be to update our models so that we don't care about our PK. We turn it into a UUID or something, and then we gain a ton of freedom. @Andrew-Dickinson is going to try doing this.

WillNilges:
https://pypi.org/project/shortuuid/

Andrew-Dickinson:
Jk lol. They're in the DB but they're still not on the map. Should be a pretty easy map tweak to get them to show up

Andrew-Dickinson:
Fixed via #571

Issue #275, opened 2024-03-25
Add ability to assign "vanity" NNs


Andrew-Dickinson:
For some hub sites, we like having the ability to specify a _specific_ available NN instead of the lowest available.

Some examples are grouping nearby hubs/large buildings with similar NNs or matching the address number of the building

This might be as simple as a runbook for admins for how to manually assign NNs, however an "NN browser" tool which lets you see the available NNs and pick one seems like it would be pretty cool

Andrew-Dickinson:
#506 makes this possible via the admin UI when creating a `Node` object from scratch, but I'll leave this issue open to track the ability to do this via the standard form

PR #506, opened 2024-09-03
Implement UUID based IDs for all models

Andrew-Dickinson:
Implements UUID based IDs for every single model in the database. This is primarily to enable the presence of `Node` objects that don't consume network number allocations (see #489).

The biggest changes are to the `Install` and `Node` models, as these currently use fields we care about as their primary key (`install_number`  and `network_number`). We need to keep these fields around, of course, and we need to be careful about enforcing constraints on them now that they are not protected by Django on our behalf. We set them both to `unique=True`, and we add special logic to each:

In the case of `install_number`, we use a custom SQL migration to instruct postgres to mark the field as an [identity column](https://www.postgresql.org/docs/17/ddl-identity-columns.html). This allows the database to manage the auto-increment behavior (which helps avoid race conditions and simplifies logic on the python side). We do have to make a few modifications to the internal workings of `Install.save()` in order to get Django to understand that we shouldn't pass a value for this field to Postgres in most cases.

In the case of `network_number` I was able to re-use a lot of the existing auto assignment logic, but just needed to gate things so that the auto-assignment only occurs when the node status is `Active` (so that we don't accidentally give out NNs to planned nodes when they are saved for other reasons). This seems to work remarkably well, it makes it impossible to have an `Active` node with no NN, for example.

For both fields I also added protection logic to prevent the modification of the number once set, so they are "immutable fields" from our python code's perspective

With these model changes we were able to get rid of a lot of the non-standard config I had done to change the postgres column names (yay!).

These changes also made `network_number` nullable, so we have to go through every place it is referenced and add safety checks, usually just by filtering out any nodes that don't have it. The largest changes here are in an overhauled handler for the NN assignment form. It needs to be able to handle a ton of edge cases around pre-existing nodes with null NNs, so I decided to restructure it around the new logic rather than spagetti-ing on top of the existing structure.

I modified the import script to take advantage of the newly nullable nature of `network_number` in order to import potential hubs with a `Planned` `Node` object (and no NN). As well, as some small tweaks to member de-duplication in order to accomodate uuid based primary keys.

The API endpoints changed a bit, primarily we now support two different styles for access to `Node` and `Install` objects. URLs can either contain the primary key UUID, or the corresponding integer identifier. Django apparently has OOTB support for this, we just have to add separate URL entries and views for each style and it can infer the datatype automatically when routing to our views. Also added a bunch of new filters to the lookup endpoints to handle API clients that want to search by either NN/Install# or ID.

I also did a little cleanup of some admin UI performance, editing, and display issues, and modified the "Add Node" page so that now network numbers can be manually specified (if the volunteer doesn't use the new node form at least), so it's arguable that this PR also completes #275. Also added some tests to continue to expand the admin UI test coverage

The migrations for this were absolutely atrocious and took me an entire afternoon. This is because Django (and to some degree postgres) doesn't actually support modifying the datatypes of your primary keys, so we have to do lots of "copy the data somewhere else while we modify the schema and then put it back". I really don't want to consolidate them, they're a pain to test, and I don't think having them expanded is hurting anything

Aside from the migrations, the one big papercut I ran into is that we loose the ability to click on the APs in the admin UI to navigate the appropriate device objects (since the hacky trick we had been using was based on integer IDs for these objects). I think I'm prepared to let that just be limitation going forward, as the map stuff is pretty hacky already and I can't think of a way to cram a UUID into a reasonably sized integer reversibly. 

Closes: #489 
Closes: #499


Andrew-Dickinson:
TODO: Fix bulk import and update wiki page

src/meshapi/serializers/model_api.py
```diff
@@ -1,56 +1,95 @@
...
+
     buildings: serializers.PrimaryKeyRelatedField = serializers.PrimaryKeyRelatedField(many=True, read_only=True)
     devices: serializers.PrimaryKeyRelatedField = serializers.PrimaryKeyRelatedField(many=True, read_only=True)
+    installs = serializers.SerializerMethodField()
+
+
+class NodeEditSerializer(NodeSerializer):
+    class Meta:
+        model = Node
+        fields = "__all__"
```
Andrew-Dickinson, line 80: Is this not inherited from the parent class?

src/meshapi/static/widgets/auto_populate_location.js
```diff
@@ -15,14 +15,16 @@ $(document).ready(function($) {
...
+        if (e.target.id === "id_Building_nodes-0-building") {
+            searchId = e.target.value;
+            if (e.target.value) {
+                $('#location_autocomplete_button').removeClass("disabled-button")
+            } else {
+                $('#location_autocomplete_button').addClass("disabled-button")
+            }
         }
-    });
+    })
```
WillNilges, line 27:
```suggestion
    });
```

src/meshapi/views/forms.py
```diff
@@ -333,7 +335,8 @@ class Meta:
     "NNFormSuccessResponse",
     fields={
         "detail": serializers.CharField(),
-        "building_id": serializers.IntegerField(),
+        "building_id": serializers.UUIDField(),
+        "install_id": serializers.UUIDField(),
```
WillNilges, line 339: Note to self: maybe we should make this consistent with the other serializers.

src/meshapi/views/forms.py
```diff
@@ -395,29 +398,40 @@ def network_number_assignment(request: Request) -> Response:
...
-                "detail": message,
-                "building_id": nn_install.building.id,
-                "install_number": nn_install.install_number,
-                "network_number": nn_install.node.network_number,
-                "created": False,
-            },
-            status=status.HTTP_200_OK,
-        )
+    # Track if we have made any changes, so we know what status code to return
+    dirty = False
```
WillNilges, line 402: I wonder if it would be possible to use django's `.original` or something at the end to check if anything changed, and just always try to save. Would that be simpler than a dirty bit?

src/meshapi/views/map.py
```diff
@@ -122,7 +122,7 @@ def list(self, request: Request, *args: List[Any], **kwargs: Dict[str, Any]) ->
             ap_json = {
                 # Hacky, but we have no choice, we need this to present as a "node" object to the
                 # map frontend and not conflict with any existing installs
-                "id": 1_000_000 + ap.id,
+                "id": 1_000_000 + (ap.id.int % 1_000_000),
```
WillNilges, line 125: this is compatibility with the old map. That's fine, I guess lol

src/meshapi/admin/inlines.py
```diff
@@ -70,21 +72,30 @@ class NonrelatedBuildingInline(BetterNonrelatedInline):
     fields = ["primary_node", "bin", "street_address", "city", "zip_code"]
     readonly_fields = fields
 
-    add_button = True
+    add_button = False
```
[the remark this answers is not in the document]
Andrew-Dickinson: We don't want this button (discussed in person)

src/meshapi/tests/test_nn.py
```diff
@@ -37,6 +37,7 @@ def setUp(self):
         self.building.save()
 
         inst = sample_install.copy()
+        inst["status"] = Install.InstallStatus.REQUEST_RECEIVED
```
Andrew-Dickinson, line 40: Add test that install numbers that’s used for their own node are set to pending, not NN assigned

src/meshapi/migrations/util/custom_many_to_many.py
```diff
@@ -1,6 +1,8 @@
 """
 !!!!
 
+NB: This file is no longer used, but needs to be maintained for the sake of migrations not having issues
+
```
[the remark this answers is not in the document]
Andrew-Dickinson: Yes. I guess I can make that explicit in this comment

src/meshapi/models/building.py
```diff
@@ -1,14 +1,15 @@
...
 from django.contrib.postgres.fields import ArrayField
 from django.core.validators import MinValueValidator
 from django.db import models
+from django.db.models import ManyToManyField
 from django_jsonform.models.fields import ArrayField as JSONFormArrayField
 
 from meshdb.utils.spreadsheet_import.building.constants import AddressTruthSource
 
 from .node import Node
-from .util.custom_many_to_many import CustomColumnNameManyToManyField
```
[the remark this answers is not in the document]
Andrew-Dickinson: Yes!

src/meshapi/serializers/model_api.py
Andrew-Dickinson: It is!

src/meshapi/models/install.py
```diff
@@ -15,17 +27,17 @@ class InstallStatus(models.TextChoices):
...
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
     # Install Number (generated when form is submitted)
-    install_number = models.AutoField(
-        primary_key=True,
-        db_column="install_number",
-    )
+    # We applied a custom migration to mark this column as "GENERATED BY DEFAULT AS IDENTITY"
+    # in the SQL DDL which causes this field to autopopulate at the DB level
+    install_number = models.IntegerField(unique=True, null=False, editable=False)
```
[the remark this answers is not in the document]
Andrew-Dickinson: https://forum.djangoproject.com/t/django-4-2-is-a-2nd-autofield-like-field-on-a-model-possible/28278

This man is our savior 🙏 

I thought this was impossible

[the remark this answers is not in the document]
Andrew-Dickinson: I think this will work now

src/meshapi/models/node.py
```diff
@@ -91,19 +102,55 @@ class NodeType(models.TextChoices):
...
+                if original.network_number is not None and self.network_number != original.network_number:
+                    raise ValidationError("Network number is immutable once set")
+
+            if self.network_number:
+                pre_existing_node_with_nn = Node.objects.filter(network_number=self.network_number).first()
+                if pre_existing_node_with_nn is not None and pre_existing_node_with_nn != self:
+                    raise ValidationError("Network number already in use by another node")
+
+                if not TYPE_CHECKING:
+                    Install = apps.get_model(app_label="meshapi", model_name="Install")  # noqa: F811
```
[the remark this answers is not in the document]
Andrew-Dickinson: Done

src/meshapi/models/node.py
```diff
@@ -91,19 +102,55 @@ class NodeType(models.TextChoices):
...
             with transaction.atomic():
                 with advisory_lock("nn_assignment_lock"):
                     self.network_number = get_next_available_network_number()
                     super().save(*args, **kwargs)
         else:
+            if not self._state.adding:
+                original = Node.objects.get(pk=self.pk)
+                if original.network_number is not None and self.network_number != original.network_number:
+                    raise ValidationError("Network number is immutable once set")
+
```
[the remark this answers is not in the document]
Andrew-Dickinson: This was a good callout, I was able to identify a few opportunities for deduplicaiton and refactoring and it looks much better now

src/meshapi/views/map.py
Andrew-Dickinson: 🫡

src/meshapi/views/forms.py
Andrew-Dickinson: We could do this, but it would require an extra DB call for each model we are saving, to fetch the pre-edited copy. Django does not keep track of the original on our behalf unfortunately.

I don't think those extra DB calls are that bad from a performance perspective (we're doing a ton of DB calls here anyway). But I think they represent a step up in complexity from the dirty bit approach. Especially when you consider the edge cases around the objects not necessarily existing, etc. 

If you feel strongly we could do it, there might be a robustness argument for it

scripts/import_spreadsheet_dump.sh
```diff
@@ -43,8 +43,8 @@ do
 done
 
 # Fix the auto numbering sequence for installs
-max_install_number=$(($($DOCKER_PG_COMMAND -c "SELECT MAX(install_number) FROM $table_name" -At) + 1))
-$DOCKER_PG_COMMAND -c "ALTER SEQUENCE "$table_name"_install_number_seq RESTART WITH $max_install_number"
+max_install_number=$(($(${DOCKER_PG_COMMAND} -c "SELECT MAX(install_number) FROM meshapi_install" -At) + 1))
+${DOCKER_PG_COMMAND} -c "ALTER SEQUENCE meshapi_install_install_number_seq RESTART WITH ${max_install_number}"
```
WillNilges, line 47: Hardcoding the variables?

src/meshapi/models/node.py
```diff
@@ -103,35 +106,37 @@ class NodeType(models.TextChoices):
 
     def save(self, *args: Any, **kwargs: Any) -> None:
         if not self.network_number and self.status == self.NodeStatus.ACTIVE:
+            # If the Node has been set to active (or newly created as active) but
```
WillNilges, line 109: thanks for the comments <3

src/meshapi/util/network_number.py
```diff
@@ -10,6 +13,54 @@ def no_op() -> None:
...
 
 
+def validate_network_number_unused_and_claim_install_if_needed(
+    network_number_candidate: int, pre_existing_node_id: Optional[uuid.UUID]
+) -> None:
+    """
+    Helper function to encapsulate the validation logic for the attachment of an NN
+    to a new node. We need to validate that the NN is unused as either an install number
+    or network number on an existing node. However, there are some edge cases, like if
+    this node is being
```
WillNilges, line 23: Looks like this got cut off here?

src/meshdb/settings.py
```diff
@@ -248,6 +248,11 @@
     ],
     "DEFAULT_SCHEMA_CLASS": "drf_spectacular.openapi.AutoSchema",
     "DEFAULT_FILTER_BACKENDS": ["django_filters.rest_framework.DjangoFilterBackend"],
+    "DEFAULT_RENDERER_CLASSES": [
+        "rest_framework.renderers.JSONRenderer",
+        # Removes HTML form (doesn't work with NestedKeyObjectRelatedField)
+        "meshapi.util.drf_renderer.OnlyRawBrowsableAPIRenderer",
```
WillNilges, line 254: What does this change? Does this change the look of API endpoints in the web browser? Can you show me an example?

WillNilges (approved):
A couple of questions.

scripts/import_spreadsheet_dump.sh
Andrew-Dickinson: What do you mean? The only thing we need to update is the install number sequence, which has a predictable name

src/meshdb/settings.py
Andrew-Dickinson: Sent you a screenshot on slack. This removes the HTML form from the bottom of the page in the browser based API view. The JSON one is still there.

scripts/import_spreadsheet_dump.sh
WillNilges: Gotcha. I was a bit confused because this is used as a variable in the for loop down below. Just wanted to check.

src/meshapi/util/network_number.py
Andrew-Dickinson: Fixed thanks

scripts/import_spreadsheet_dump.sh
Andrew-Dickinson: Note: there is no longer a for loop down below. This is the only sequence that needs updating, since the PKs are all UUIDs now

PR #508, opened 2024-09-05
Reset dev branch from prod backup daily + other changes to backups

WillNilges:
This PR adds a celery task to reset the dev database from a prod backup at midnight and stops taking backups from dev. The `prod1` S3 Bucket config is now hardcoded in `settings.py`, because `dbbackup` doesn't seem to support specifying an S3 path on the fly. There might be some way to do it, but this was easier and I figured we weren't getting much value out of having the backups on dev run. We should be able to monitor our backups with Datadog.

This PR now also adds a maintenance mode and a place for us to put custom permissions.


WillNilges:
Codecov made me feel bad so I added a few tests. Should I be extra and add tests for my `get_most_recent_object` function and maybe `scramble_members` or should I leave it?

This PR is also another one that makes me wish we had a staging branch I could deploy to first, but thankfully we haven't launched yet so I'm still allowed to break things.

WillNilges:
This also reminds me that I need to see if I can figure out what we can learn from the redis integrations.

src/meshapi/management/commands/scramble_members.py
```diff
@@ -25,6 +25,9 @@ class Command(BaseCommand):
     help = "Updates all members with fake name, email, and phone number. Clears notes."
 
     def add_arguments(self, parser: ArgumentParser) -> None:
+        parser.add_argument(
+            "--noinput", action="store_true", help="Tells Django to NOT prompt the user for input of any kind."
+        )
```
Andrew-Dickinson, line 30: This is useful for the import script also, ty

src/meshapi/tasks.py
```diff
@@ -12,21 +13,61 @@
...
 from meshdb.celery import app as celery_app
+from meshdb.settings import DBBACKUP_STORAGE_OPTIONS
 
 
 @celery_app.task
-def run_database_backup() -> None:
-    logging.info("Running database backup task")
+def run_database_backup() -> bool:
+    # Don't run a backup unless it's prod1
+    environment = f"\"{os.environ.get('MESHDB_ENVIRONMENT', '')}\""
```
Andrew-Dickinson, line 22: Should we make `MESHDB_ENVIRONMENT` a variable in `settings.py`?

src/meshapi/tasks.py
```diff
@@ -12,21 +13,61 @@
...
+    if os.environ.get("MESHDB_ENVIRONMENT") != "prod1":
```
Andrew-Dickinson, line 23: We probably should put the env names into a constant somewhere

src/meshapi/tasks.py
```diff
@@ -12,21 +13,61 @@
...
         raise e
 
+    return True
+
+
+@celery_app.task
+def reset_dev_database() -> bool:
+    # Only reset dev environments (very important!)
+    environment = f"\"{os.environ.get('MESHDB_ENVIRONMENT', '')}\""
+    if "dev" not in os.environ.get("MESHDB_ENVIRONMENT", ""):
```
Andrew-Dickinson, line 45: Any chance there's a way to do this at the celery level? Like only schedule certain tasks under certain conditions? Not a big deal but would be pretty clean to have like a map of envs and list the workers we want for each

src/meshapi/tasks.py
```diff
@@ -12,21 +13,61 @@
...
     except Exception as e:
         logging.exception(e)
         raise e
 
+    return True
+
+
+@celery_app.task
+def reset_dev_database() -> bool:
+    # Only reset dev environments (very important!)
```
Andrew-Dickinson, line 43: This definitely isn't going to nuke my local setup if I run the celery worker, right?

src/meshapi/tasks.py
```diff
@@ -12,21 +13,61 @@
...
+    environment = f"\"{os.environ.get('MESHDB_ENVIRONMENT', '')}\""
+    if "dev" not in os.environ.get("MESHDB_ENVIRONMENT", ""):
+        logging.warn(f"Not resetting this database. This environment is: {environment}")
+        return False
+
+    logging.info(f"Running database reset task. This environment is: {environment}")
+
+    if not os.environ.get("AWS_ACCESS_KEY_ID") or not os.environ.get("AWS_SECRET_ACCESS_KEY"):
+        logging.error("Could not run backup. Missing AWS credentials!")
```
Andrew-Dickinson, line 52:
```suggestion
        logging.error("Could not run database reset. Missing AWS credentials!")
```

src/meshapi/tasks.py
```diff
@@ -12,21 +13,61 @@
...
+    try:
+        latest_backup = task_utils.get_most_recent_object(
+            DBBACKUP_STORAGE_OPTIONS["bucket_name"], DBBACKUP_STORAGE_OPTIONS["location"]
+        )
+
+        if not latest_backup:
+            raise ValueError("Could not get most recent dbbackup while resetting dev environment.")
+
+        management.call_command("dbrestore", "--noinput", "-i", latest_backup)
+        management.call_command("scramble_members", "--noinput")
```
Andrew-Dickinson, line 64: Do we need to consider the security implications of a bad guy having dev access and getting de-anonymized data in between when these commands run?

src/meshapi/util/task_utils.py
```diff
@@ -0,0 +1,21 @@
...
+# Returns the key to the most recently modified object in an S3 bucket
+def get_most_recent_object(bucket_name: str, prefix: str) -> str | None:
+    s3_client = boto3.client("s3")
+    objects = s3_client.list_objects_v2(Bucket=bucket_name, Prefix=prefix)
+
+    if "Contents" not in objects:
+        logging.error(f"No objects found in '{bucket_name}/{prefix}'.")
+        return None
+
+    most_recent_object = max(objects["Contents"], key=lambda obj: obj["LastModified"])
```
Andrew-Dickinson, line 15: Hmm this isn't going to work, S3 is paginated. You have to loop through an iterator given to you by `boto3` to hit all the objects if you want to do this, but that's pretty inefficient. A more efficient solution would be to include the date in the backup prefix, like: `s3://meshdb-data-backups/meshdb-backups/prod1/2024/09/06/default-meshdb-celery-59496585dd-75vqq-2024-08-30-230003.psql.bin`. Then you can just list todays backups using the prefix and guarantee it will fit on one page

src/meshapi/tasks.py
WillNilges: Not a bad idea. Definitely less verbose. What if we do that for all of our environment variables?

src/meshapi/tasks.py
WillNilges: I disagree. This is directly derived from the github action that built the helm chart. It would just make one more step for creating a new environment for no reason. Plus it could be useful for local dev purposes.

src/meshapi/tasks.py
WillNilges: Yeah I looked for that briefly. If you insist, I can keep looking. Maybe I could conditionally declare the beats.

src/meshapi/tasks.py
WillNilges: Not unless your environment matches `dev`.

src/meshapi/tasks.py
WillNilges: further proof that I need to go to the optometrist.

src/meshapi/tasks.py
WillNilges: Hmmm, I wonder how practical it would be to have some kind of sandbox to do the scramble in, or swap between a primary and secondary database. It would add some complexity... Alternatively we can maybe kick everyone out while this job runs.

src/meshapi/tasks.py
Andrew-Dickinson: I'll counter that it's not obvious to a dev what the env names are and an enum listing all of them would make questions like "is `dev` for beta deployments or local dev?" much more obvious. Maybe a wiki page would be good?

src/meshapi/util/task_utils.py
WillNilges: Shit. I knew there was gonna be some issue like that. Unfortunately I have zero control over precisely how dbbackup stores backups in S3. We're still on my fork of it, so maybe I could patch it. Is there a way to ask boto for a specific time range or something?

src/meshapi/tasks.py
Andrew-Dickinson: Ah I see, you edited the template to add a default of `local`

src/meshapi/util/task_utils.py
Andrew-Dickinson: Nope, that's a limitation of S3, it's a flat store where all the info you care about needs to be encoded into the object key

Andrew-Dickinson: I suppose one thing you have going for you is that I set the bucket to delete backups older than 90 days, so this will be a constant number of objects rather than continuously increasing

Andrew-Dickinson: I think it actually might work if you set `prefix` to `s3://meshdb-data-backups/meshdb-backups/prod1/default-meshdb-celery-59496585dd-75vqq-YYYY-MM-DD`. Files are not special in S3, they're just a part of the key

src/meshapi/tasks.py
Andrew-Dickinson: Yeah that may be a simpler solution. I don't really like the complexity of having to hot-swap postgres databases, but I bet django supports it to at least some degree

src/meshapi/tasks.py
WillNilges: We should definitely make a wiki page. I'll think about making an enum.

src/meshapi/tasks.py
WillNilges: Yeah, I would like to not do all that nonsense just to ensure security. I might look into logging people out.

src/meshapi/util/task_utils.py
WillNilges: Problem with that is `meshdb-celery-59496585dd-75vqq` is a pod name, which is non-deterministic, ephemeral, and (almost) impossible to get (especially when the most recent backup may have happened 59 minutes ago). If I could somehow specify a wildcard with a date, that could work.

src/meshapi/tasks.py
Andrew-Dickinson: Even that may be trickier than you expect, since you’ll need to invalidate all existing sessions, as well as prevent the creation of new ones

src/meshapi/util/task_utils.py
Andrew-Dickinson: No wildcard support :( just prefixes

src/meshapi/tasks.py
WillNilges: Done. Not going to do it for all env vars in this PR but that would be an ez first issue lol

WillNilges: https://github.com/nycmeshnet/meshdb/issues/515

src/meshapi/tasks.py
WillNilges: Stubbed one out: https://wiki.mesh.nycmesh.net/books/6-services-software/page/environments

src/meshapi/tasks.py
WillNilges: Done. I will leave the checks and the unit tests in place _just in case_

src/meshapi/util/task_utils.py
WillNilges: Ah, wait a sec: https://django-dbbackup.readthedocs.io/en/stable/configuration.html#dbbackup-filename-template

Andrew-Dickinson: The forward slashes would make the console UI nicer, but that's the only difference. Doesn't impact performance or anything

WillNilges: oh, also, if you don't specify a backup file, the latest one gets used: https://github.com/jazzband/django-dbbackup/blob/master/dbbackup/management/commands/_base.py#L111

¯\\\_(ツ)_/¯

src/meshapi/tasks.py
WillNilges: Ah, I can create maintenance mode middleware: https://medium.com/@keagileageek/putting-a-django-site-in-maintenance-mode-creating-a-django-maintenance-mode-middleware-7c5474f62491

WillNilges: I have done it

but at what cost?

src/meshapi/models/permission.py
```diff
@@ -0,0 +1,22 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
+
+
+# This is an abstract model that exists only so that we can create a ContentType
+# for which can attach permissions to it.
+# There is no database table created for it
+class Permission(models.Model):
+    class Meta:
+        abstract = True
```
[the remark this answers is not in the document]
Andrew-Dickinson: I think this is the source of your test failures, so I would want to really understand why this is needed and what it does

docker-compose.yaml
```diff
@@ -83,9 +83,9 @@ services:
     expose:
       - 80
     # Useful for debugging, but should be commented out otherwise.
-    #ports:
-    #  - 80:80
-    #  - 443:443
+    ports:
+      - 80:80
+      - 443:443
```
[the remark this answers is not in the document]
WillNilges: oops that was for development

src/meshweb/views.py
```diff
@@ -33,3 +37,43 @@ def index(request: HttpRequest) -> HttpResponse:
...
+    }
+    return HttpResponse(template.render(context, request))
+
+
+# TODO (wdn): Can I make a "disable maintenance mode" button visible for admins?
+
+
+@api_view(["GET"])
+@permission_classes([HasMaintenanceModePermission])
+def enable_maintenance(request: HttpRequest) -> HttpResponse:
```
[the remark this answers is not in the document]
WillNilges: I can delete this now that I have the flags

src/meshweb/views.py
```diff
@@ -33,3 +37,43 @@ def index(request: HttpRequest) -> HttpResponse:
...
+        "message": "Please check back later.",
+        "redirect": "",
+    }
+    return HttpResponse(template.render(context, request))
+
+
+# TODO (wdn): Can I make a "disable maintenance mode" button visible for admins?
+
+
+@api_view(["GET"])
```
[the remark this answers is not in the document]
WillNilges: eat my jorts

src/meshweb/middleware.py
```diff
@@ -0,0 +1,28 @@
...
+from typing import Callable
+
+from django.http import HttpRequest, HttpResponse
+from django.shortcuts import redirect
+from django.urls import reverse
+
+MAINTENANCE_FILE = Path("/tmp/meshdb_maintenance")
+
+
+class MaintenanceModeMiddleware:
```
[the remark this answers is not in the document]
WillNilges: eat my jorts

src/meshapi/models/permission.py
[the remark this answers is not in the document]
WillNilges: This can be deleted now that I have flags

WillNilges:
oh cool we get a little UI for maintenance mode: 
image

Now, this won't be accessible when maintenance mode is on, so we'll have to keep the `maintenance/enable/` and `maintenance/disable/` views.

src/meshdb/settings.py
```diff
@@ -172,9 +188,10 @@
 # https://django-dbbackup.readthedocs.io/en/master/installation.html
 
 DBBACKUP_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
+DBBACKUP_FILENAME_TEMPLATE = "{datetime}.{extension}"
 DBBACKUP_STORAGE_OPTIONS = {
-    "bucket_name": os.environ.get("BACKUP_S3_BUCKET_NAME"),
-    "location": os.environ.get("BACKUP_S3_BASE_FOLDER"),
+    "bucket_name": "meshdb-data-backups",
```
james-otten, line 193: If these env vars are all the way gone, please remove from helm + CI

james-otten: I could go either way on continuing to reference as env vars

james-otten (approved):
For iac

Andrew-Dickinson:
I’ll look tomorrow

WillNilges:
https://wiki.mesh.nycmesh.net/books/6-services-software/page/maintenance-mode

src/meshdb/settings.py
WillNilges: I'd prefer to clean them up in a follow-up PR. I'd like to make sure everything works properly, first.

src/meshweb/views.py
```diff
@@ -33,3 +37,40 @@ def index(request: HttpRequest) -> HttpResponse:
...
+    template = loader.get_template("meshweb/maintenance.html")
+    context = {
+        "message": "Please check back later.",
+        "redirect": "",
+    }
+    return HttpResponse(template.render(context, request))
+
+
+@api_view(["POST"])
+@permission_classes([IsSuperUser])
```
Andrew-Dickinson, line 54:
```suggestion
@permission_classes([HasMaintenanceModePermission])
```
Doing it this way will have more runtime flexibility. I don't like coupling our application logic to the Django concept of superuser like that

src/meshweb/views.py
```diff
@@ -33,3 +37,40 @@ def index(request: HttpRequest) -> HttpResponse:
...
+    template = loader.get_template("meshweb/maintenance.html")
+    context = {
+        "message": "Enabled maintenance mode.",
+        "redirect": "maintenance",
+    }
+    return HttpResponse(template.render(context, request))
+
+
+@api_view(["POST"])
+@permission_classes([IsSuperUser])
```
Andrew-Dickinson, line 66:
```suggestion
@permission_classes([HasMaintenanceModePermission])
```

src/meshapi/tests/test_maintenance_mode.py
```diff
@@ -0,0 +1,89 @@
+from django.contrib.auth.models import User
+from django.test import TestCase
+from django.test.client import Client
+from flags.state import disable_flag, enable_flag, flag_disabled, flag_enabled
+
+
+class TestMaintenanceMode(TestCase):
+    c = Client()
```
Andrew-Dickinson, line 8: nit: you can just use the built in `self.client` for this

src/meshapi/tests/test_tasks.py
```diff
@@ -0,0 +1,58 @@
...
+class TestTasks(TestCase):
+    def setUp(self):
+        pass
+
+    @mock.patch("django.core.management.call_command")
+    @mock.patch("meshapi.tasks.MESHDB_ENVIRONMENT", "prod1")
+    def test_run_database_backup(self, mock_call_command_func):
+        os.environ["AWS_ACCESS_KEY_ID"] = "fake"
+        os.environ["AWS_SECRET_ACCESS_KEY"] = "alsofake"
+        self.assertTrue(run_database_backup())
```
Andrew-Dickinson, line 20: nit: If you want to avoid returning a value from `run_database_backup()` you can do `mock_call_command_func.assert_called_once()`

src/meshapi/management/commands/create_groups.py
```diff
@@ -46,5 +46,7 @@ def handle(self, *args: Any, **options: Any) -> None:
                 or act == "view"
                 or obj in ["user", "token", "tokenproxy", "celeryserialzerhook"]
                 or code == "assign_nn"
+                or code == "update_panoramas"
+                or code == "maintenance_mode"
```
Andrew-Dickinson, line 50:
```suggestion
```
If we want this perm to be "superuser only"

src/meshapi/tests/test_tasks.py
WillNilges: Yeah but I want to know that it was successful. I suppose instead I could raise an exception, that would work.

src/meshapi/tests/test_maintenance_mode.py
WillNilges: damn that would have been good to know all the way back last august.

Andrew-Dickinson: Yeah… I only noticed this a few weeks ago when looking at an example in the docs

Andrew-Dickinson (approved):

Andrew-Dickinson:
FWIW I'm okay with overriding codecov in this case, 88% is close enough

src/meshdb/settings.py
james-otten: https://github.com/nycmeshnet/meshdb/issues/535

PR #517, opened 2024-09-08
Load Ticket Number from OSTicket during import process

### LINKED ISSUE #510: Backfill OSTicket IDs
We should add a module to the import script which backfills the OSTicket IDs by dumping from OSTicket and parsing the text for install number

Andrew-Dickinson:
OSTicket has a CSV export feature that we can (somewhat hackily) tap into to bulk export ticket data. It also has a "node number" metadata column. This lets us join the install number into the ticket number.

Also renamed `ticket_id` -> `ticket_number` project-wide since this is what OSTicket calls the value that is displayed in the UI. There is also an ID field in OSTicket which is different, so not doing this rename would create some confusion I think

Closes #510


src/meshdb/utils/spreadsheet_import/fetch_osticket_data.py
```diff
@@ -0,0 +1,148 @@
...
+logger.configure()
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "meshdb.settings")
+django.setup()
+
+from meshapi.models import Install
+
+
+def download_export(session, export_id):
+    download_response = session.get("https://support.nycmesh.net/scp/export.php?id=" + export_id)
```
WillNilges, line 23: should we parameterize this somewhere?

WillNilges (approved):

src/meshdb/utils/spreadsheet_import/fetch_osticket_data.py
Andrew-Dickinson: The spreadsheet import stuff needs to get killed with fire, so probably not

PR #531, opened 2024-09-09
Add support for links between AP dots

### LINKED ISSUE #514: Can't add links between AP dots
Access point devices can't be selected in the from/to device `select2` dropdowns in the `Add Link` page

They also aren't imported automatically at spreadsheet import time, but I bet we can do this by search the device notes field

Andrew-Dickinson:
Adds support for links to AP dots on the map. First, we make sure that they are imported from the spreadsheet. Next, we ensure that all device types can be selected in the "edit link" detail page. Finally, we modify the map endpoint to include these links with the appropriate "node IDs" (this is a bit hacky since our devices don't cleanly map to anything it considers a "node" and we just use some UUID math to make sure it doesn't collide with any install or network numbers)

Closes #514


Andrew-Dickinson:
Marking as draft so I remember not to merge until after #506 goes

WillNilges (approved):

PR #533, opened 2024-09-12
Bump meshweb to 3 replicas

WillNilges:
It would be good to have 3 replicas of Meshweb running for HA purposes and to make deployments a little smoother. I have seen moments of unavailability on dev before that I would like to avoid. This is now possible because MeshDB is stateless.


james-otten (approved):

Andrew-Dickinson (approved):
Only one copy of the celery workers, right?

WillNilges:
yes

PR #536, opened 2024-09-15
Enable tracing for meshdb individually

WillNilges:
[Single-step instrumentation](https://docs.datadoghq.com/tracing/trace_collection/automatic_instrumentation/single-step-apm/?tab=kubernetes#specifying-at-the-service-level) is still a bit imprecise, so I'm going to switch back to specifying per-service tracing for now.

### DIFF
--- a/infra/helm/meshdb/templates/meshweb.yaml
+++ b/infra/helm/meshdb/templates/meshweb.yaml
@@ -13,7 +13,7 @@ spec:
   template:
     metadata:
       annotations:
-        admission.datadoghq.com/python-lib.version: v2.11.2
+        admission.datadoghq.com/python-lib.version: v2.12
       {{- with .Values.podAnnotations }}
         {{- toYaml . | nindent 8 }}
       {{- end }}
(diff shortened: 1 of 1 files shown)


infra/helm/meshdb/templates/meshweb.yaml
```diff
@@ -12,8 +12,9 @@ spec:
       {{- include "meshdb.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
       annotations:
+        admission.datadoghq.com/python-lib.version: v2.11.2
```
james-otten, line 16: Not quite latest, but very recent

james-otten (approved):

infra/helm/meshdb/templates/meshweb.yaml
WillNilges: Oh I think I could just pin it to a version tag: https://hub.docker.com/r/datadog/dd-lib-python-init/tags

Same way we do with our python dependencies.

WillNilges:
```suggestion
        admission.datadoghq.com/python-lib.version: v2.12
```

PR #538, opened 2024-09-16
[celery] Change limits, tweak readiness, and convert to statefulset

WillNilges:
What it says on the tin. I wanna do this to ensure that there's one and only one celery pod at a time.

https://stackoverflow.com/questions/68074742/can-a-deployment-ensure-that-there-is-never-more-than-one-pod-running

https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

### DIFF
--- a/infra/helm/meshdb/charts/celery/templates/deployment.yaml
+++ b/infra/helm/meshdb/charts/celery/templates/deployment.yaml
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: StatefulSet 
+kind: StatefulSet
 metadata:
   name: {{ include "celery.fullname" . }}
   namespace: meshdb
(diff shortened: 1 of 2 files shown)


infra/helm/meshdb/charts/celery/templates/deployment.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: StatefulSet
```
james-otten, line 2: Can you remove the trailing space

james-otten (approved):

infra/helm/meshdb/charts/celery/templates/deployment.yaml
WillNilges:
```suggestion
kind: StatefulSet
```

PR #542, opened 2024-09-17
Use SQL explorer

WillNilges:
This PR adds a nice SQL explorer directly to the github project. I have set up a read-only SQL Role for use with it, added a permission and a group.

https://github.com/explorerhq/sql-explorer/

- [x] create group(?)
- [x] Make readonly role that only has access to meshdb data


src/meshapi/management/commands/create_groups.py
WillNilges, line 1: This is actually kinda cursed and we could totally do this with a migration. Maybe for a later PR.

src/meshapi/migrations/0031_create_meshdb_ro.py
```diff
@@ -0,0 +1,87 @@
+# Generated by Django 4.2.15 on 2024-09-18 01:53
```
Andrew-Dickinson, line 1: To be clear: I do support doing this sort of thing with a migration, but I'll note that in #506 you had me take my custom SQL migration and move it into the main application code so that we don't have to carefully extract it when we nuke and re-generate the migrations prior to launch.

I think this is not that bad to extract, but we will have to do it

src/meshapi/tests/test_views_explorer.py
```diff
@@ -0,0 +1,28 @@
+from django.contrib.auth.models import Group, User
+from django.test import TestCase
+
+from meshapi.tests.group_helpers import create_groups
+
+
+class TestViewsGetExplorer(TestCase):
```
Andrew-Dickinson, line 7: Should we add a case for an unauthenticated user? 

Also maybe a simple check that something like `SELECT * FROM meshapi_members` gives HTTP 200? Just to make sure we catch any regressions of the permissions error you were getting at the start of this feature. I could imagine someone adding a new table and forgetting to exclude it in `settings.py` for example. Would be good to give them a red X on their PR

Andrew-Dickinson (approved):
Approving, but we really should add a slightly more detailed test

src/meshapi/migrations/0031_create_meshdb_ro.py
WillNilges: Can you jog my memory about waht specific migration this was? Was it this? 

https://github.com/nycmeshnet/meshdb/pull/506/files/89603f171137062468c8e06d89523d754f1d2c0d#diff-543b583d981a7222ba8ac3f5136152fe92754cb704a303e21c55fc59d742cd69

Andrew-Dickinson: This the migration in question:
https://github.com/nycmeshnet/meshdb/pull/506/files#diff-64acc684bdd1b6f78227e5b50f104b3a7fb2bc79aec6ea69eff92678583512acR11

and this is the line that gets us out of needing it to be present:
https://github.com/nycmeshnet/meshdb/pull/506/commits/14d558b623d718d9b944b9f92c0062f8e4e5f594#diff-297eadf0b00ee217b77ccf35aff6ebc6a5b398624387be526b67d9bf9fbb054eR21

Andrew-Dickinson: This the migration in question:
https://github.com/nycmeshnet/meshdb/pull/506/files#diff-64acc684bdd1b6f78227e5b50f104b3a7fb2bc79aec6ea69eff92678583512acR11

and this is the line that gets us out of needing it to be present:
https://github.com/nycmeshnet/meshdb/pull/506/commits/14d558b623d718d9b944b9f92c0062f8e4e5f594#diff-297eadf0b00ee217b77ccf35aff6ebc6a5b398624387be526b67d9bf9fbb054eR21

Andrew-Dickinson: Yeah I think a `manage.py` command, is really our only other option, and to be clear, I think this is the right place to do this. It's just evidence that we shouldn't generally plan to mess with migrations in the future

src/meshapi/tests/test_views_explorer.py
WillNilges: Getting some weird-ass errors when I try to test this:
```

Traceback (most recent call last):
  File "/home/wilnil/Code/nycmeshnet/meshdb/.venv/lib/python3.12/site-packages/asgiref/local.py", line 89, in _lock_storage
    asyncio.get_running_loop()
RuntimeError: no running event loop

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/wilnil/Code/nycmeshnet/meshdb/.venv/lib/python3.12/site-packages/django/utils/connection.py", line 58, in __getitem__
    return getattr(self._connections, alias)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wilnil/Code/nycmeshnet/meshdb/.venv/lib/python3.12/site-packages/asgiref/local.py", line 118, in __getattr__
    return getattr(storage, key)
           ^^^^^^^^^^^^^^^^^^^^^
AttributeError: '_thread._local' object has no attribute 'r'
```

For what it's worth, the library has some pretty extensive tests of its own: https://github.com/explorerhq/sql-explorer/blob/master/explorer/tests/test_views.py

WillNilges: OK, because we're doing raw sql query shit, the tests for this end up positively awful because Django is so goddamn insistent on its concept of a `test_*` database. I have spent hours fighting with the tooling and all I have accomplished is breaking the everliving shit out of everything.

WillNilges: OK, there, I got it: https://docs.djangoproject.com/en/5.1/topics/testing/advanced/#testing-primary-replica-configurations

don't make me do anything else.

WillNilges: ope and even though it #worksonmymachine it doesn't work in github for some alien reason

Andrew-Dickinson: Yeah it’s crashing quite spectacularly

PR #556, opened 2024-09-18
Make name dedupe logic case-insensitive & prefer longer names

### LINKED ISSUE #545: Don't add notes for duplicate names if the only difference is case

Andrew-Dickinson:
Closes #545


src/meshdb/utils/spreadsheet_import/parse_member.py
```diff
@@ -70,10 +72,20 @@ def merge_member_objects(members_and_installs: List[Tuple[Member, List[int]]]) -
         if merged_member.name is None:
             merged_member.name = member.name
         else:
-            if merged_member.name != member.name and member.name:
+            if merged_member.name.lower() in member.name.lower() and len(merged_member.name) < len(member.name):
+                # If they gave us a longer name on a successive submission, and the earlier submission
```
WillNilges, line 76: I guess we're deduping on emails, so it's fine, but it feels wack to me to have this rule in a world where `John Smith` and `Johnny Smith` both exist.

WillNilges (approved):

src/meshdb/utils/spreadsheet_import/parse_member.py
Andrew-Dickinson: Well in that case we would actually take the first one that ocurrs. We only take the longer name if it's a strict superset of the shorter name. So `John` could become `John smith`, but `Johnny` would be maintained. This is more conservative than we have to be but helps out in a surprising number of cases

PR #562, opened 2024-09-19
Change ticket number to charfield

Andrew-Dickinson:
Turns out OSTIcket ticket numbers can have leading zeros and those leading zeros are important to URLs not 404ing 

This PR documents that and converts the field to a charfield so we don't drop them


src/meshapi/migrations/0031_alter_install_ticket_number.py
```diff
@@ -0,0 +1,21 @@
...
+        ("meshapi", "0030_merge_20240916_1938"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="install",
+            name="ticket_number",
+            field=models.CharField(
+                blank=True,
+                help_text="The ticket number of the OSTicket used to track communications with the member about this install. Note that although this appears to be an integer, it is not. Leading zerosare important, so this should be stored as a string at all times",
```
WillNilges, line 17:
```suggestion
                help_text="The ticket number of the OSTicket used to track communications with the member about this install. Note that although this appears to be an integer, it is not. Leading zeros are important, so this should be stored as a string at all times",
```

src/meshapi/models/install.py
```diff
@@ -53,10 +53,12 @@ class InstallStatus(models.TextChoices):
     )
 
     # OSTicket Ticket Number
-    ticket_number = models.IntegerField(
+    ticket_number = models.CharField(
         blank=True,
         null=True,
-        help_text="The ticket number of the OSTicket used to track communications with the member about this install",
+        help_text="The ticket number of the OSTicket used to track communications with the member about "
+        "this install. Note that although this appears to be an integer, it is not. Leading zeros"
```
WillNilges, line 60:
```suggestion
        "this install. Note that although this appears to be an integer, it is not. Leading zeros "
```

src/meshapi/tests/test_install.py
```diff
@@ -103,6 +103,14 @@ def test_update_install_unset_install_number(self):
         self.assertEqual(install.id, uuid.UUID("23ef170c-f37d-44e3-aaac-93dae636c86e"))
         self.assertEqual(install.install_number, 45)
 
+    def test_osticket_number_leading_zeros(self):
```
WillNilges, line 106: Can/should we add a testcase to reproduce the issue?

WillNilges (approved):
A few massive nits, but :shipit: 

That's funny as fuck

src/meshapi/tests/test_install.py
Andrew-Dickinson: That's what this is, right? Like the current issue is that leading zeros aren't tracked. And this test confirms that they are

PR #566, opened 2024-09-19
Add LOS objects to permissions script

### LINKED ISSUE #568: Admins don't have permissions to add webhooks
Whoops, you need superuser to do that, we should add that to the groups script

Andrew-Dickinson:
@WillNilges do we need to do anything special to run this in prod or does it do it automatically?

Closes #568


WillNilges:
We can do it at startup @Andrew-Dickinson 
https://github.com/nycmeshnet/meshdb/pull/581

WillNilges (approved):

PR #576, opened 2024-09-20
Fix: admin map nav broken for buildings, links, devices & loses

### LINKED ISSUE #560: LOSes & Links are not shown in the map view
When viewing a LOS or link in the map view, the installs on either side of the link are not shown in the admin map view, it is blank

Andrew-Dickinson:
Fixes #560 

Forgot to update this stuff in the UUID overhaul (#506)


Andrew-Dickinson:
Hmm we can't easily do a UX test without some heavyweight stuff (DDog, Selenium, etc.). It might be good to have JS unit tests though. Let me see if there's a relatively painless way to do that

Andrew-Dickinson:
Okay so it turns out there is not. Trying to test browser JS code in Node really sucks, and I'm not good enough at javascript to do it without stacking crappy workarounds several layers deep. No tests for you

WillNilges:
womp womp

PR #581, opened 2024-09-20
Create groups on startup

WillNilges:
This will make sure that groups are always up to date on startup

(TODO: Make group creation a migration)

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -18,3 +18,5 @@ python manage.py migrate
 echo 'Collecting Static Files...'
 python manage.py collectstatic --no-input
 
+echo 'Creating Groups...'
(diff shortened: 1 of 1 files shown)


Andrew-Dickinson (approved):

scripts/entrypoint.sh
```diff
@@ -18,3 +18,5 @@ python manage.py migrate
 echo 'Collecting Static Files...'
 python manage.py collectstatic --no-input
 
+echo 'Creating Groups...'
+python maange.py create_groups
```
Andrew-Dickinson, line 22:
```suggestion
python manage.py create_groups
```

WillNilges: that's what I get for doing shit on the train

Issue #547, opened 2024-09-18
Disable pano editing in building admin UI


Andrew-Dickinson:
Willard says make it a feature flag

PR #589, opened 2024-09-22
Add feature flag for panoramas (and other fixes)

WillNilges:
This changes the `EDIT_PANORAMAS` env var to a feature flag, so that it can be toggled live.

This also updates the default values of environment variables so that you don't have to change them when switching between containerized dev and baremetal dev.

### DIFF
--- a/src/meshapi/tests/test_admin_change_view.py
+++ b/src/meshapi/tests/test_admin_change_view.py
@@ -5,6 +5,7 @@ from typing import Dict, Optional
 import bs4
 from django.contrib.auth.models import Group, User
 from django.test import Client, TestCase
+from flags.state import enable_flag
 from rest_framework.authtoken.models import TokenProxy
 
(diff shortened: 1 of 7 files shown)


WillNilges:
Closes #547

src/meshapi/tests/test_admin_change_view.py
```diff
@@ -464,9 +465,11 @@ def test_add_new_planned_node_no_nn(self):
         self.assertEqual(node.notes, "Test notes")
 
     def test_add_building(self):
+        enable_flag("EDIT_PANORAMAS")
```
Andrew-Dickinson, line 468: It might make more sense to change these tests to avoid editing the panos, since this is the state prod will be in, right?

Andrew-Dickinson (approved):

src/meshapi/tests/test_admin_change_view.py
WillNilges: Good call, added a test to add a building with the panorama editor disabled 👍🏻

PR #601, opened 2024-09-23
Fix: exports get captured by captive nav and don't download

### LINKED ISSUE #597: Exports don't work
Admin UI exports don't work, probably because we're capturing them in the admin form capture logic from #486

Exports should be excluded from this capture logic so that the download works

Andrew-Dickinson:
Whoops, we broke file downloads when we captured form based nav in #486, this adds an exception for file download so that these are not intercepted

Fixes #597


WillNilges (approved):
Any chance we can get tests on this?

Andrew-Dickinson:
Absolutely not, already went down that road in #576, doing tests for code like this really really sucks

PR #604, opened 2024-09-24
Fix: NYC addresses are sent to random New England towns

Andrew-Dickinson:
This has been a persisent import script issue, but it seemed to only happen for badly formatted addresses so I never looked that carefully into it. However today I noticed it was happening for perfectly valid YAYB addresses, and decided to try to fix it. I think this is gonna make things a great deal better


Andrew-Dickinson:
Merging as this is import script related and Willard is still on vacation

Issue #564, opened 2024-09-19
Nodes where the member moves out but leaves the device up are imported as inactive


Andrew-Dickinson:
Do a scan pass at the end of the import script which marks nodes active even if they have no active installs, if they have active connected devices

Andrew-Dickinson:
Examples: NN525, NN236

PR #606, opened 2024-09-24
Fix: nodes without active installs are hidden on the map

Andrew-Dickinson:
This filter dates back all the way to before we had a node object, and the following code is all capable of handling the "only inactive installs" case. It's blocking the display of the nodes we fixed in #564 (like NN238 and NN525), so it seems we should remove it


src/meshapi/views/map.py
```diff
@@ -69,7 +69,7 @@ def get_queryset(self) -> List[Install]:  # type: ignore[override]
         # include the nns we haven't already covered via install num
         covered_nns = {install.install_number for install in all_installs}
         for node in (
-            Node.objects.filter(~Q(status=Node.NodeStatus.INACTIVE) & Q(installs__status__in=ALLOWED_INSTALL_STATUSES))
```
WillNilges, line 72: It looks like `ALLOWED_INSTALL_STATUSES` is only mentioned here and on L216. Should we remove it there as well? And if so, should we remove the variable entirely?

WillNilges (approved):

src/meshapi/views/map.py
Andrew-Dickinson: Good catch, I'll get rid of this entirely

PR #609, opened 2024-09-24
Ask user to confirm info before creating Install

WillNilges:
Closes https://github.com/nycmeshnet/meshforms/issues/56


src/meshapi/views/forms.py
```diff
@@ -134,6 +137,51 @@
             {"detail": "Your address could not be validated."}, status=status.HTTP_500_INTERNAL_SERVER_ERROR
         )
 
+    changed_info: dict[str, str | int] = {}
+
+    # TODO: Notify member if we changed any of their information
+    # Name (won't touch), email (won't touch), phone, st addr, unit (won't touch), city, State, Zip
+    if formatted_phone_number and r.phone != formatted_phone_number:
+        logging.warning(f"Changed phone_number: {formatted_phone_number} != {r.phone}")
```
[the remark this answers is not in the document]
WillNilges: sick AI bro you're AI is all over the place

WillNilges:
Idea: Instead of logging PII, add some marker to the join form log expressing that there was a problem. That would require appending to an existing S3 object in meshforms, since it would save the submission immediately upon entry.

WillNilges:
Yeah i think i already thought of this but http code

WillNilges:
I think now all's I gotta do is update Meshforms. Having just spent 30+ hours learning more React, I have i d e a s.

WillNilges:
Maybe I could include the PII in the Notes of the created user as well...?

src/meshapi/views/forms.py
```diff
@@ -134,6 +138,50 @@ def join_form(request: Request) -> Response:
...
+        changed_info["phone"] = formatted_phone_number
+
+    if r.street_address != nyc_addr_info.street_address:
+        logging.warning(f"Changed street_address: {r.street_address} != {nyc_addr_info.street_address}")
+        changed_info["street_address"] = nyc_addr_info.street_address
+
+    if r.city != nyc_addr_info.city:
```
Andrew-Dickinson, line 153: Can the ZIP or state mismatch? Or are we excluding that possibility already above?

src/meshapi/views/forms.py
```diff
@@ -134,6 +138,50 @@ def join_form(request: Request) -> Response:
             {"detail": "Your address could not be validated."}, status=status.HTTP_500_INTERNAL_SERVER_ERROR
         )
 
+    changed_info: dict[str, str | int] = {}
+
+    # TODO: Notify member if we changed any of their information
+    # Name (won't touch), email (won't touch), phone, st addr, unit (won't touch), city, State, Zip
+    if formatted_phone_number and r.phone != formatted_phone_number:
```
Andrew-Dickinson, line 145: Hmm so if they type `(555) 555-5555` and we reformat it to `+1 555-555-5555`, they will get a pop-up warning them we've made a change? Perhaps we should use a frontend validation library to format the phone number into a consistent format to prevent this

WillNilges:
Maybe Django is formatting it after the fact? IDK I need to do some more testing.

src/meshapi/views/forms.py
WillNilges: We're still gonna want to do some validation in the backend, I think. I'm trying to get the behavior to be consistent. See my comments from this morning.

I mean, maybe we just silently accept whatever formatting our backend does. The main concern here is the address.

WillNilges:
```
>>> print(phonenumbers.format_number(phonenumbers.parse("+1 614-307-4056", "US"), phonenumbers.PhoneNumberFormat.INTERNATIONAL))
+1 614-307-4056
>>> print(phonenumbers.format_number(phonenumbers.parse("+1614-307-4056", "US"), phonenumbers.PhoneNumberFormat.INTERNATIONAL))
+1 614-307-4056
```

image

src/meshapi/tests/test_join_form.py
```diff
@@ -961,7 +961,7 @@ def test_member_moved_and_used_non_primary_email_join_form(self):
             f"but got {response3.status_code}.\n Response is: {response3.content.decode('utf-8')}",
         )
 
-        validate_successful_join_form_submission(self, "Valid Join Form", s, response3, expected_member_count=2)
+        validate_successful_join_form_submission(self, "Valid Join Form", s, response3, expected_member_count=3)
```
WillNilges, line 964: It's 5AM so I'm making a wild fucking guess, but this should be making 3 instead of 2 now, since phone number is no longer deduped

src/meshapi/views/forms.py
WillNilges: We now have frontend validation for the phone number. It's still validated in the backend, but this should not be a concern.

src/meshapi/views/forms.py
WillNilges: They totally can, but I think I have a test for it now, (at least with NJ state but NYC address). I think it's acceptable to have this validation in the backend with the rest of our address code, but maybe we should have some explicit check for it so we can return a more specific error. Right now, it just tells the member that Non-NYC registrations are not accepted.

src/meshapi/widgets.py
```diff
@@ -15,7 +15,7 @@ class PanoramaViewer(JSONFormWidget):
     def __init__(self, schema: dict):
         super().__init__(schema)
 
-    def pano_get_context(self, name: str, value: str) -> dict:
+    def pano_get_context(self, name: str, value: str = "[]") -> dict:
```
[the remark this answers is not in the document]
WillNilges: I think this is some git fuckery because I definitely didn't write this in this PR.

src/meshapi/views/query_api.py
```diff
@@ -75,7 +75,7 @@ class Meta:
         parameters=[query_form_password_param],
         summary="Query & filter based on Install attributes. "
         "Results are returned as flattened spreadsheet row style output",
-        auth=[],
+        auth=["bearer"],
```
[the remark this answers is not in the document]
WillNilges: I think this is some git fuckery because I definitely didn't write this.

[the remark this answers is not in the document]
WillNilges: I think this is some git fuckery because I definitely didn't write this.

src/meshapi/views/forms.py
```diff
@@ -134,6 +138,50 @@ def join_form(request: Request) -> Response:
             {"detail": "Your address could not be validated."}, status=status.HTTP_500_INTERNAL_SERVER_ERROR
         )
 
+    # Keep a blank JoinFormRequest to mutate with the info we've joined. We'll send it back
```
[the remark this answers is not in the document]
WillNilges: Nope. We only touch Phone or Address if it was mutated by our validation. We don't do dupe checking because I believe the current opinion is that it would be dangerous to let people know that info exists (doxxing purposes etc)

WillNilges:
yolo'ing it with approval to do so from @Andrew-Dickinson

Issue #343, opened 2024-04-18
DB Encryption


james-otten:
Data in postgres needs to be encrypted at rest

Andrew-Dickinson:
Do we also need to consider all backups?

james-otten:
```
Volumes:
  pg-data-vol:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  meshdb-postgres-encyrpted-pvc
```

```
pvc-1aa61cba-9d65-4701-8681-807d71c5fd81   20Gi       RWO            Delete           Bound    meshdb/meshdb-postgres-encyrpted-pvc   longhorn-encrypted   <unset>                          96m
```

PR #616, opened 2024-09-28
Complete volume swap

james-otten:
Complete the volume swap to an encrypted volume for postgres.

I will take care of this, approve but **do not merge**.

Resolves https://github.com/nycmeshnet/meshdb/issues/343


Andrew-Dickinson (approved):
What does the migration look like? Will we just need to restore from a backup?

james-otten:
> What does the migration look like? Will we just need to restore from a backup?

That could be done with just this PR and it would be less complicated.

1. Scale everything but postgres to zero
2. Execute a manual backup
3. Merge this
4. Scale back to zero
5. Restore backup
6. Scale back up (or just re-deploy)

PR #621, opened 2024-09-29
Only backup 'default'

james-otten:
Only try to backup the "default" database, ignoring "readonly" which would be a subset of "default".

Resolves error:
```
pg_dump: error: query failed: ERROR:  permission denied for table django_migrations
```

Docs:
https://django-dbbackup.readthedocs.io/en/master/configuration.html#dbbackup-databases


WillNilges (approved):
Good catch

PR #640, opened 2024-10-11
Resource request and limit

james-otten:
Set resource requests + limits based on current production usage as seen in el doggo.

Docs: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/


Andrew-Dickinson (approved):

infra/helm/meshdb/values.yaml
```diff
@@ -19,17 +19,13 @@ pg:
...
-  # choice for the user. This also increases chances charts run on environments with little
-  # resources, such as Minikube. If you do want to specify resources, uncomment the following
-  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
-  # limits:
-  #   cpu: 100m
-  #   memory: 128Mi
-  # requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  resources:
```
WillNilges, line 22: We should definitely pin Postgres to a core. It deserves it.

infra/helm/meshdb/values.yaml
```diff
@@ -47,7 +43,13 @@ meshweb:
...
   podSecurityContext: {}
   securityContext: {}
-  resources: {}
+  resources:
+    limits:
+      cpu: 1
+      memory: 512Mi
+    requests:
+      cpu: 200m
+      memory: 350Mi
```
WillNilges, line 52: Pin this guy to 512Mi. 200m is _probably_ fine as a request, as I think most of the CPU is on startup with the tracer.

infra/helm/meshdb/values.yaml
```diff
@@ -56,7 +58,13 @@ nginx:
...
   podSecurityContext: {}
   securityContext: {}
-  resources: {}
+  resources:
+    limits:
+      cpu: 1
+      memory: 100Mi
+    requests:
+      cpu: 250m
+      memory: 30Mi
```
WillNilges, line 67: This guy _definitely_ doesn't need more than 250mcore and like, 128Mi of RAM. If our web proxy is taking up more than that, we've either got wayyyy more users than we thought we did, or we have a performance problem.

infra/helm/meshdb/values.yaml
```diff
@@ -83,7 +91,13 @@ redis:
...
   podSecurityContext: {}
   securityContext: {}
-  resources: {}
+  resources:
+    limits:
+      cpu: 1
+      memory: 512Mi
+    requests:
+      cpu: 50m
+      memory: 30Mi
```
WillNilges, line 100: Redis _really_ doesn't like not being pinned, especially as an in-memory cache. I'd say pin it to 256Mi and like..... 250mcore. If that doesn't work, bump it.

infra/helm/meshdb/values.yaml
```diff
@@ -92,7 +106,13 @@ pelias:
...
   podSecurityContext: {}
   securityContext: {}
-  resources: {}
+  resources:
+    limits:
+      cpu: 1
+      memory: 512Mi
+    requests:
+      cpu: 250m
+      memory: 250Mi
```
WillNilges, line 115: No fucking idea here. Can probably get away with pinning limit to the smaller request here tho.

WillNilges (changes requested):
Unless you have a very good reason, such as an app is startup-hungry, we should be pinning requests and limits. It makes things more predictable, simpler to debug, and there's no reason not to if we know we have the resources to spare.

Each of these pins should be done one at a time.

(Addendum, this is all based on my own experience. If you have a Datadog graph that proves me wrong, I'd like to see it.)

james-otten:
> Can we do these one at a time? It'll be easier to debug if it's causing startup issues or something.
> 
> Also, I don't think we have the ability to gate our prod releases on whether or not dev is in a good state, which makes changes like these a little spooky 👻

Very spooky, has me saying AAAAHHHHHHH

1. https://github.com/nycmeshnet/meshdb/pull/655
2. https://github.com/nycmeshnet/meshdb/pull/656
3. https://github.com/nycmeshnet/meshdb/pull/657
4. https://github.com/nycmeshnet/meshdb/pull/658
5. https://github.com/nycmeshnet/meshdb/pull/659

james-otten:
Several others

PR #661, opened 2024-10-27
Resource

james-otten:
- Expect meshweb to have a bit more CPU during startup
- Allow  pelias more memory to be used during startup, which is then freed during normal usage


WillNilges (approved):
Can you show data to back this change up?

james-otten:
> Can you show data to back this change up?

Slack

PR #668, opened 2024-10-31
Add ability to replay join records

WillNilges:
This isn't quite done, but the general idea is there.

It adds a command you can use to replay join records. You'll have to exec into a pod to do it right now, but I can probs expose this via a web interface at some point.

It has a basic mock test, but I need to make it more robust before I trust it.


WillNilges:
image

WillNilges:
https://github.com/nycmeshnet/meshforms/pull/93

WillNilges:
Had some thoughts on the train. I should include install number as the indicator or success.

WillNilges:
I would like to specify more rigid types and ensure parity between the join record format in JS and Python. They both read and write from the bucket so it's imperative that they're the same.

Also, I haven't addressed the problem that we're going to be getting a _lot_ of these so the list will grow quite large.

WillNilges:
Theoretically the bucket can grow quite large. Maybe one thing I could do is try to improve the way I fetch the failed submissions.

Then again, computers are fast, and the sub-optimal code might not matter until much, much later.

WillNilges:
why is codecov mad I literally fixed it it literally has the wrong commit.

src/meshapi/util/join_records.py
```diff
@@ -0,0 +1,106 @@
...
+    def __init__(self, data: dict[str, JoinRecord]) -> None:
+        self.bucket_name: str = "mock_bucket"
+        # Store join record by S3 key and value.
+        self.bucket: dict[str, JoinRecord] = data
+
+    def upload(self, join_record: JoinRecord, key: str) -> None:
+        self.bucket[key] = join_record
+
+    def get_all(self) -> list[JoinRecord]:
+        return list(self.bucket.values())
```
Andrew-Dickinson, line 106: nit: should this mock live in a test file?

src/meshapi/util/join_records.py
```diff
@@ -0,0 +1,106 @@
...
+    @abstractmethod
+    def get_all(self) -> list[JoinRecord]:
+        pass
+
+
+class JoinRecordProcessor(JoinRecordProcessorInterface):
+    def __init__(self) -> None:
+        self.s3_client = boto3.client(
+            "s3",
+            endpoint_url=JOIN_RECORD_ENDPOINT,
```
Andrew-Dickinson, line 62: If this env var isn't provided, we should probably make this `None` so we let boto grab the default, it's kinda annoying to have to set this manually for real S3

src/meshapi/management/commands/replay_join_records.py
```diff
@@ -0,0 +1,74 @@
...
+from prettytable import PrettyTable
+
+from meshapi.util.join_records import JOIN_RECORD_BASE_NAME, JoinRecord, JoinRecordProcessor
+from meshapi.views.forms import JoinFormRequest, process_join_form
+
+
+class Command(BaseCommand):
+    help = "Replay join form submissions that we may not have accepted properly"
+
+    def add_arguments(self, parser: ArgumentParser) -> None:
```
Andrew-Dickinson, line 16: Perhaps for efficiency and damage control reasons it would be good to have a flag to filter submissions by time? (and default it to like -1 week from the current time)

src/meshapi/validation.py
```diff
@@ -19,6 +19,9 @@
 DOB_BUILDING_HEIGHT_API_URL = "https://data.cityofnewyork.us/resource/qb5r-6dgf.json"
 
 
+# FIXME (wdn): When we can't reach the internet to get the email, this returns 400.
+# That's wrong. It should be 500 because it's our fault. I've seen this reject
+# valid emails because of this and it makes testing offline impossible unless you mock it.
```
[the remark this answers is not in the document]
WillNilges: oops, yep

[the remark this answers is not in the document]
WillNilges: OK deleting the TODO and just using this?

https://github.com/nycmeshnet/meshdb/issues/692

src/meshapi/util/join_records.py
WillNilges: We could do that. It is kinda nice for the mock to live right next to the thing that it mocks.

src/meshapi/validation.py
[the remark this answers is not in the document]
WillNilges: I linked the issue.

WillNilges:
Added an issue for the post-launch stuff we need to do: https://github.com/nycmeshnet/meshforms/issues/102

WillNilges:
i got rid of the mock, it was causing more problems than it was solving so now this is an integration test. Sue me and die.

src/meshapi/management/commands/replay_join_records.py
WillNilges: Captured in https://github.com/nycmeshnet/meshdb/issues/699

docker-compose.yaml
```diff
@@ -81,10 +80,39 @@ services:
...
+  createbuckets:
+    image: minio/mc
+    depends_on:
+      - minio
+    entrypoint: >
+      /bin/sh -c "
+      /usr/bin/mc config host add myminio http://minio:9000 sampleaccesskey samplesecretkey;
+      /usr/bin/mc mb myminio/meshdb-join-form-log;
+      exit 0;
+      "
```
Andrew-Dickinson, line 109: Would it be better to use the AWS CLI here instead? Like you did in the join form PR

src/meshapi/util/join_records.py
```diff
@@ -0,0 +1,99 @@
...
+import boto3
+from botocore.client import ClientError, Config
+
+from meshapi.views.forms import JoinFormRequest
+
+JOIN_RECORD_ENDPOINT = os.environ.get("S3_ENDPOINT", None)
+JOIN_RECORD_BUCKET_NAME = os.environ.get("JOIN_RECORD_BUCKET_NAME")
+JOIN_RECORD_PREFIX = os.environ.get("JOIN_RECORD_PREFIX", "sample-basename")
+JOIN_RECORD_ACCESS_KEY = os.environ.get("AWS_ACCESS_KEY_ID")
+JOIN_RECORD_SECRET_KEY = os.environ.get("AWS_SECRET_KEY")
```
Andrew-Dickinson, line 18: remove unused?

src/meshapi/management/commands/replay_join_records.py
```diff
@@ -0,0 +1,75 @@
...
+        parser.add_argument(
+            "--noinput", action="store_true", help="Tells Django to NOT prompt the user for input of any kind."
+        )
+
+        parser.add_argument("--all", action="store_true", help="Fetch all Join Records, not just failed ones")
+
+        parser.add_argument("--look", action="store_true", help="Print Join Records and quit")
```
Andrew-Dickinson, line 23: nit: This might be safer as the inverse. i.e. a flag you pass in order to enable writes, not a flag to disable them

I see we have a confirmation prompt though so probably fine

src/meshapi/tests/test_replay_join_records.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+# Integration test to ensure that we can fetch JoinRecords from an S3 bucket,
+# replay them into the join form, and insert the objects we expect.
+# We should test a happy case, a case when the JoinRecord is bad (like a JoinRecord
+# somehow from Russia), and maybe mock a MeshDB 500 that just ends in us putting
+# the data back.
+class TestReplayJoinRecords(TestCase):
+    def setUp(self) -> None:
+        p = JoinRecordProcessor()
```
Andrew-Dickinson, line 19: nit: if you made this an instance variable you wouldn't need to construct so many of them

src/meshapi/tests/test_replay_join_records.py
```diff
@@ -0,0 +1,145 @@
...
+        p.flush_test_data()
+
+    def tearDown(self) -> None:
+        p = JoinRecordProcessor()
+        p.flush_test_data()
+
+    # This is just to make codecov happy
+    def test_s3_content_to_join_record(self):
+        sample_key = "dev-join-form-submissions/2024/11/01/11/33/49.json"
+        sample_s3_content = """{"first_name":"Jon","last_name":"Smith","email_address":"js@gmail.com","phone_number":"+1 585-475-2411","street_address":"197 Prospect Place","apartment":"1","city":"Brooklyn","state":"NY","zip_code":"11238","roof_access":true,"referral":"I googled it.","ncl":true,"trust_me_bro":false,"code":"201","replayed":0,"install_number":1002}"""
```
Andrew-Dickinson, line 29: nit: Might be nice to let the auto-formatter have a crack at this JSON blob by making it a python dict wrapped by `json.dumps()`

Andrew-Dickinson (approved):

docker-compose.yaml
WillNilges: Not sure if I could get the entrypoint to behave properly. This is just for development, so it's fine.

WillNilges:
https://github.com/nycmeshnet/meshdb/issues/699

PR #680, opened 2024-11-02
Add slack secret to celery

james-otten:


Andrew-Dickinson (approved):
Wow good catch. This would have taken me a long time

PR #682, opened 2024-11-02
Set HSTS to 1 year

### LINKED ISSUE #642: Increase HSTS timeout to actually enforce it for a reasonable duration
In `settings.py`, set the HSTS timeout to much higher based on industry standards (probably 31536000?)

james-otten:
I haven't seen any issues with HTTP traffic. Resolves #642

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security

### DIFF
--- a/src/meshdb/settings.py
+++ b/src/meshdb/settings.py
@@ -52,7 +52,7 @@ FLAGS: Dict[str, Any] = {
 
 USE_X_FORWARDED_HOST = True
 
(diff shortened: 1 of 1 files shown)


james-otten:
> Scary. What data do we have to confirm this is safe?

Network tab in devtools

Andrew-Dickinson (approved):
Fair enough, I suppose we haven't had any issues with 30 seconds

PR #703, opened 2024-11-09
Add devices names to link.__str__ for "self links"

Andrew-Dickinson:
This may be a bit controversial since it makes things a bit bulky at times, but looking at the string representations for "self links" in slack is very confusing.

`NN7489 -> NN7489` is so much more confusing than `nycmesh-7489-epr7 -> nycmesh-7489-north`

For now, we only include the device names for self-links (links between devices within a single node)

This heterogenous format does look a bit messy at times, but results in a lot more clarity for single links in isolation

Screenshot 2024-11-09 at 18 48 45

Also switches to a double headed arrow to indicate that links are not directional

### DIFF
--- a/src/meshapi/models/link.py
+++ b/src/meshapi/models/link.py
@@ -78,8 +78,22 @@ class Link(models.Model):
     )
 
     def __str__(self) -> str:
-        if self.from_device.node.network_number and self.to_device.node.network_number:
-            return f"NN{self.from_device.node.network_number} → NN{self.to_device.node.network_number}"
+        from_name = None
+        to_name = None
+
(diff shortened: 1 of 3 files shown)


WillNilges:
I would be OK with changing it to the device name everywhere, since all devices ought to have the NN in the name anyway.

WillNilges (approved):
See my above comment

Andrew-Dickinson:
You're sure that isn't too verbose? They're so long....

Andrew-Dickinson:
Let's keep it this way for now, and if we decide we don't like the inconsistency later we can tweak it

PR #705, opened 2024-11-11
Add captcha validation

### LINKED ISSUE #78: Add a CAPCHA for the join form
This is going to be an open API into our database. We should protect it from spam. From my limited experience, there are several captchas out there that are pretty plug-and-play.

Andrew-Dickinson:
Adds captcha validation to the join form backend. Supports two different token types, the first is an implicit score based on the user's behavior. If that token scores low, then we will respond with HTTP 401 and the frontend will prompt them to do a checkbox captcha, which will generate a second token.

See also https://github.com/nycmeshnet/meshforms/pull/108

Closes #78


src/meshapi/views/forms.py
```diff
@@ -105,6 +113,21 @@ def join_form(request: Request) -> Response:
 
 
 def process_join_form(r: JoinFormRequest, request: Optional[Request] = None) -> Response:
```
Andrew-Dickinson, line 115: Does this belong here, maybe it should it be included in `join_form()` above so that we don't need to worry about this validation in the join form replay CLI?

src/meshapi/views/forms.py
```diff
@@ -101,6 +109,21 @@ def join_form(request: Request) -> Response:
         logging.exception("TypeError while processing JoinForm")
         return Response({"detail": "Got incomplete form request"}, status=status.HTTP_400_BAD_REQUEST)
 
+    if not settings.DEBUG and not DISABLE_RECAPTCHA_VALIDATION:
```
Andrew-Dickinson, line 112: Does this belong here? I put it here so that we don't need to worry about this validation in the join form replay CLI, but does that open us up to any risks?

.env.sample
```diff
@@ -73,3 +73,10 @@ OSTICKET_NEW_TICKET_ENDPOINT=https://devsupport.nycmesh.net/api/http.php/tickets
 
 SLACK_ADMIN_NOTIFICATIONS_WEBHOOK_URL=
 SLACK_JOIN_REQUESTS_CHANNEL_WEBHOOK_URL=
+
+# Captcha checking is automatically disabled when DEBUG=True,
+# so this whole section is fine to leave unchanged for local dev
+RECAPTCHA_DISABLE_VALIDATION=False
+RECAPTCHA_SERVER_SECRET_KEY_V2=
+RECAPTCHA_SERVER_SECRET_KEY_V3=
+RECAPTCHA_INVISIBLE_TOKEN_SCORE_THRESHOLD=0.5
```
Andrew-Dickinson, line 82: Need to add these to IaC

src/meshapi/views/forms.py
```diff
@@ -101,6 +109,21 @@ def join_form(request: Request) -> Response:
...
+        try:
+            request_source_ip, request_source_ip_is_routable = get_client_ip(request)
+            if not request_source_ip_is_routable:
+                request_source_ip = None
```
Andrew-Dickinson, line 116: Add logging statement here to check that our reverse proxy stuff is working right

src/meshapi/views/forms.py
```diff
@@ -49,6 +55,8 @@ class JoinFormRequest:
     referral: str
     ncl: bool
     trust_me_bro: bool  # Used to override member data correction
+    recaptcha_invisible_token: Optional[str]
+    recaptcha_checkbox_token: Optional[str]
```
Andrew-Dickinson, line 59: Do these via headers instead

src/meshapi/views/forms.py
Andrew-Dickinson: See https://github.com/nycmeshnet/meshdb/pull/712

src/meshapi/validation.py
```diff
@@ -204,3 +210,57 @@ def geocode_nyc_address(street_address: str, city: str, state: str, zip_code: st
...
+        raise ValueError("Invalid recaptcha token")
+
+    # If there is no score in the response, we are dealing with a v2 token,
+    # which has no concept of score, only binary success/failure of the manual
+    # checkbox captcha. In this case, we can be confident this is a human since
+    # they have completed the checkbox captcha, so we return 1.0 (100% human score)
+    return response_json.get("score", 1.0)
+
+
+def validate_captcha_tokens(
```
WillNilges, line 240: Nit: `validate_captcha_tokens` or `validate_recaptcha_tokens`?

src/meshapi/views/forms.py
```diff
@@ -101,6 +107,29 @@ def join_form(request: Request) -> Response:
...
+            if recaptcha_checkbox_token == "":
+                recaptcha_checkbox_token = None
+
+            validate_captcha_tokens(
+                recaptcha_invisible_token,
+                recaptcha_checkbox_token,
+                request_source_ip,
+            )
+        except Exception:
+            logging.exception("Captcha validation failed")
```
WillNilges, line 130: Should we maybe print the exception here just in case it's not one of the "usual" ones?

WillNilges (approved):

src/meshapi/views/forms.py
Andrew-Dickinson: `logging.exception` does this automatically

src/meshapi/validation.py
Andrew-Dickinson: Done

PR #707, opened 2024-11-18
Add Stats Graph SVG Render View

Andrew-Dickinson:
Adds an endpoint to render SVG files for the website to display, in the style of the current [nycmesh.net/stats](https://nycmesh.net/stats) page

Also does a bit of refactoring and cleanup in meshweb, we should probably spread the views out into multiple files, and we don't need to use the DRF decorators for non-API views

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ dependencies = [
     "django-sql-explorer==5.2.*",
     "django-simple-history==3.7.*",
     "prettytable==3.11.*",
+    "matplotlib==3.9.*",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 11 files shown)


james-otten (approved):

Andrew-Dickinson:
The new graphs aren't identical, but they're really close. The biggest thing that's missing is the little green % diff indicator. I don't think matplotlib can really draw that, but if anyone has tips I would appreciate them
Screenshot 2024-11-17 at 23 55 14

Screenshot 2024-11-17 at 23 54 49

Andrew-Dickinson:
Okay changed it so that we just render the graph in MeshDB. All the header stuff will be done on the website side, but I needed an API in order to expose the total counts publicly (using our standard API for this is a security hazard as the token would need to be public). Having this little JSON API also has the advantage of making it trivial to convert to a frontend graphing framework in the future if desired

WillNilges (approved):

PR #712, opened 2024-11-19
Test if IP address gathering works as expected

Andrew-Dickinson:
Don't merge


james-otten:
Closing, the real IP should come through on prod

PR #715, opened 2024-11-20
Fixes to panorama. Move files, update timeout, add retry logic

WillNilges:
- [x] Increases duration of github timeout
- [x] Adds retry logic to API calls
- [ ] ~Make Panorama Import run every 15 minutes to get a tighter feedback loop~
- [x] Add some more error handling and unit testing to address crashes
- [x] Add atomic transaction to panorama update
- [x] Change some error handling logic to ensure we bail, so that we aren't wiping out panoramas.
    - There's more I want to do here, I want to essentially ensure that we never delete, only append panoramas.


WillNilges:
very weird... we never _clear_ the panoramas, so I'm not sure why they disappear... I wanna merge this and mess with it in prod.

WillNilges:
I bet the token is wrong on that test that's 401'ing.

WillNilges:
heck
```Error: 1-21 18:34:00,304] [2884] [ERROR] Error: Got status 401 from GitHub trying to get SHA: {"message":"Bad credentials","documentation_url":"https://docs.github.com/rest","status":"401"}```

Andrew-Dickinson (approved):

PR #736, opened 2024-11-24
Add test to expose failures in Install Integrations

Andrew-Dickinson:


WillNilges:
What does "Install Integrations" mean?

WillNilges:
Is there a specific failure you're trying to expose, or is this just for coverage?

WillNilges (approved):
More tests are always good. See above comment.

Andrew-Dickinson:
> Is there a specific failure you're trying to expose, or is this just for coverage?

This is a regression test to expose the issue fixed by #734. That was causing a crash on the integrations because zip was the wrong type, and it slipped through our testing because the integrations were always disabled

PR #744, opened 2024-11-25
Deduplicate installs on join form submission when the apt matches

### LINKED ISSUE #549: Don't deduplicate installs in the join form if the install number is re-used as an NN
I was talking to Matthew Boyd tonight. For a low numbered installs that resubmit years later, we should create a new install object instead of de-duplicating if their number has been given away as an NN

Andrew-Dickinson:
I don't remember why we stopped doing this deduplication and if it was intentional. 

If there was a good reason we stopped doing it, let's keep just the tests from this PR as a way to validate our expectations about this

Closes #549


src/meshapi/views/forms.py
```diff
@@ -290,6 +290,29 @@ def process_join_form(r: JoinFormRequest, request: Optional[Request] = None) ->
...
+        building=join_form_building,
+        member=join_form_member,
+        unit__exact=r.apartment,
+        # We only recycle install objects if nothing special has happened to them,
+        # if they're not REQUEST_RECEIVED, that dramatically increases the chance that
+        # a new identifier is warranted. e.g. NN_REASSIGNED indicates that's an absolute must
+        status=Install.InstallStatus.REQUEST_RECEIVED,
+    ).first()
+    if existing_install:
+        return Response(
```
Andrew-Dickinson, line 303: Add logging statement?

WillNilges:
I'm pretty sure this was simply an oversight. It seems reasonable to me to not create a new install if the same person in the same building in the same unit fills out the join form again.

Can you walk me through what is meant by `member_filled_out_the_join_form_long_ago_and_we_recycled_their_install_number`? What does "recycling" an install number mean? Did you mean "node number"?

Andrew-Dickinson:
In this context, recycling just means the "spreadsheet row number" (which we imported as an install number due to the inactive status of that row at import time) was assigned as a network number to another install

In the sense that "node number" ~= "spreadsheet row number" I think that's correct, but it really is an install number in MeshDB parlance

WillNilges (approved):

PR #746, opened 2024-11-25
Add support for pre, post join records

WillNilges:
Fixes bugs with the command line interface for join records and implements v2 features.

TODO:
- [x] Case-by-case join record replaying
- [x] Better error handling
- [x] Tests


WillNilges:
Shit I still need to figure out what happens if we get a 409 while replaying.... 🤔

src/meshapi/util/join_records.py
```diff
@@ -98,3 +120,46 @@ def flush_test_data(self) -> None:
...
+
+        post_join_records_dict: dict[str, JoinRecord] = {}
+        for record in self.get_all(since, SubmissionStage.POST):
+            if record.version >= 2:
+                post_join_records_dict[record.uuid] = record
+
+        # For each pre-submission record, there should exist a post-submission
+        # record. If the post-submission record is missing, log a warning and add
+        # the pre-submission record to the post-submission dictionary so it is covered
+        # by the command.
```
Andrew-Dickinson, line 147: If we're checking consistency, would it be good to check that every post record has a pre record also? Not sure how we should handle that but it could theoretically happen it we probably should at least have a log statement

src/meshapi/management/commands/replay_join_records.py
```diff
@@ -64,28 +92,69 @@ def handle(self, *args: Any, **options: Any) -> None:
...
         if not options["noinput"]:
             proceed_with_replay = input("Proceed with replay? (y/N): ")
-            if proceed_with_replay.lower() != "yes" and proceed_with_replay.lower() != "y":
+            if proceed_with_replay.lower() not in ["yes", "y"]:
                 print("Operation cancelled.")
                 return
 
+        if not join_records_to_replay:
+            logging.warning("Found no join records to replay. Quitting")
+            return
```
Andrew-Dickinson, line 101: nit: should this be before the confirmation prompt? Seems annoying to have to confirm for nothing to happen

Andrew-Dickinson (approved):

PR #747, opened 2024-11-25
HTTP 500 on email validation failure

### LINKED ISSUE #692: Email Validation should return 5xx when we fail to validate the email
When we can't reach the internet to get the email, this returns 400. That's wrong. It should be 500 because it's our fault. I've seen this reject valid emails because of this and it makes testing offline impossible unless you mock it.

https://github.com/nycmeshnet/meshdb/blob/main/src/meshapi/validation.py#L22

Andrew-Dickinson:
Closes #692


src/meshapi/views/forms.py
```diff
@@ -140,8 +141,14 @@ def process_join_form(r: JoinFormRequest, request: Optional[Request] = None) ->
     if not r.email_address:
         return Response({"detail": "Must provide an email"}, status=status.HTTP_400_BAD_REQUEST)
 
-    if r.email_address and not validate_email_address(r.email_address):
-        return Response({"detail": f"{r.email_address} is not a valid email"}, status=status.HTTP_400_BAD_REQUEST)
+    try:
+        if r.email_address and not validate_email_address(r.email_address):
+            return Response({"detail": f"{r.email_address} is not a valid email"}, status=status.HTTP_400_BAD_REQUEST)
+    except EmailValidationError:
```
WillNilges, line 147: Wait, this is a little confusing, but I think I understand what's happening. Are DNSTimeoutError, SMTPCommunicationError, and TLSNegotiationError extensions of EmailValidationError?

You might consider leaving a note here for the future to point this out, if indeed that's what's happening.

src/meshapi/tests/test_join_form.py
WillNilges, line 1: Might be good to assert that an invalid email is correctly given the "benefit of the doubt" when we have an issue.

WillNilges (approved):
Two small things, but will approve to unblock. Thanks for taking care of this.

src/meshapi/views/forms.py
Andrew-Dickinson: You have it correct. Added a comment

PR #756, opened 2024-11-28
Drop revision history limit 10->4

james-otten:
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#clean-up-policy


WillNilges:
Why is this necessary? What's the problem with keeping 10?

james-otten:
Annoyance when I `get all`

WillNilges (approved):

PR #770, opened 2024-12-04
Remove phone number from 409 checking

Andrew-Dickinson:


WillNilges:
Discussed on slack yesterday. The justification is that, while we might not correctly handle all intl phone numbers, Django is validating them for US formatting only right now, and besides, an invalid phone number does nobody any good.

WillNilges (approved):

PR #773, opened 2024-12-04
fix postgres password again

WillNilges:
not verly smark from ohio etc


WillNilges:
https://github.com/nycmeshnet/meshdb/blob/main/infra/helm/meshdb/templates/secrets.yaml

james-otten (approved):

PR #776, opened 2024-12-09
Add django-admin-site-search

### LINKED ISSUE #706: Add "unified search" across all model types
Add a feature to the admin UI to search across all model types

This plugin looks promising:
https://github.com/ahmedaljawahiry/django-admin-site-search

WillNilges:
This PR implements [Django Admin Site Search](https://github.com/ahmedaljawahiry/django-admin-site-search) to MeshDB, so that you can easily search MeshDB. It adds a bar to the top of the interface you can click to open, or press `CTRL+K` (`CMD+K` on Mac) to open it.

image

The normal search bars are still present (for now) if you wish to use those.

Closes #706

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ dependencies = [
     "prettytable==3.11.*",
     "matplotlib==3.9.*",
     "django-ipware==7.0.1",
+    "django-admin-site-search==1.1.*",
 ]
 
 [project.optional-dependencies]
--- a/src/meshdb/admin.py
+++ b/src/meshdb/admin.py
@@ -1,11 +1,13 @@
-from typing import Any, List, Optional
+from typing import Any, List, Literal, Optional
 
+from admin_site_search.views import AdminSiteSearchView
 from django.contrib import admin
 from django.http import HttpRequest
-from admin_site_search.views import AdminSiteSearchView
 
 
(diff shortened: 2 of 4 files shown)


WillNilges:
There are some issues I've spotted with this, namely sometimes, the results are duplicated in the UI for whatever reason. It took me 10 minutes to install this and it's really useful, so I think this is a bug worth living with. Given a few days with the developer, I might be able to figure it out.

image

Andrew-Dickinson:
Some of the duplicates are due to model inheritance, but weird that the devices section is copied whole

Andrew-Dickinson (approved):

Andrew-Dickinson:
Is there a way to do queryset filtering the way we do on the regular search so that we don’t give the admins links to the device objects underneath sectors and APs?

also I don’t think this incorporates the custom ranking logic, does it? That was very important for small length queries (I.e. “3”). The results for that were totally useless prior to the ranking changes

WillNilges:
I'll have to get back to you on that. As I understood (at 2AM last night), the mode I've switched it to should be doing that filtering and using the same search logic (but I didn't test it that thoroughly).

I'm pretty sure there are ways to have custom logic. I will see if I can avoid reinventing the wheel, and if not, maybe I'll open another issue with the maintainer.

PR #777, opened 2024-12-09
Add tracing to Map endpoints

WillNilges:
I'd like to add some Datadog tracing to the map endpoints so we can understand if Andrew's theory of very heavy Python objects is correct. This will give us better viz in APM.

I will probably also enable profiling: https://docs.datadoghq.com/profiler/enabling/python/


WillNilges:
https://docs.datadoghq.com/getting_started/tracing/

PR #784, opened 2024-12-12
Fix csp

james-otten:
It should be just that subdomain, not additional subdomains.


Andrew-Dickinson (approved):
You've confirmed this works?

src/meshdb/settings.py
```diff
@@ -82,7 +82,7 @@
     "https://cdn.redoc.ly",  # Redoc
     "blob:",  # Redoc
     "'unsafe-inline'",  # TODO: Remove me https://github.com/nycmeshnet/meshdb/issues/645
-    "*.browser-intake-us5-datadoghq.com",
+    "browser-intake-us5-datadoghq.com",
```
james-otten, line 85:
```suggestion
    "https://browser-intake-us5-datadoghq.com",
```

PR #794, opened 2024-12-24
Add "Fiber Connected Hub" and Map Display Override Options

Andrew-Dickinson:
Adds a new hub type "Fiber Connected Hub" which renders on the map as a big blue "supernode" dot

Also adds the ability to override the display of links:
Screenshot 2024-12-24 at 17 05 10

Includes changes from #793


src/meshapi/migrations/0005_historicallink_map_display_override_and_more.py
WillNilges, line 1: Can you squash these migrations?

src/meshapi/serializers/map.py
```diff
@@ -80,13 +80,17 @@ def get_synthetic_notes(self, install: Install) -> Optional[str]:
...
         # In the case of multiple dots per node, we only want to
         # make the one that actually corresponds to the NN the big dot (the "fake" install)
         # for the real install numbers that don't match the network number, leave them as red dots
         if install.node.type != Node.NodeType.STANDARD and self._is_node_dot(install):
-            synthetic_notes.append(install.node.type)
+            if install.node.type == Node.NodeType.FIBER_HUB:
+                # Draw fiber-connected hubs with the big blue "supernode" dot
+                synthetic_notes.append(Node.NodeType.SUPERNODE)
+            else:
+                synthetic_notes.append(install.node.type)
```
WillNilges, line 93: Is this because... the map doesn't support FIBER_HUB?

src/meshapi/serializers/map.py
```diff
@@ -149,6 +153,17 @@ class Meta:
     installDate = JavascriptDateField(source="install_date")
 
     def convert_status_to_spreadsheet_status(self, link: Link) -> str:
+        if link.map_display_override:
+            map_override_to_spreadsheet_status: Dict[str, str] = {
```
WillNilges, line 157: 📓 Note: What happens if we update the MapDisplayOverride and forget to update this? Is there a test that evokes that?

WillNilges (commented):
Is there a slack thread about this?

Andrew-Dickinson:
Should we do this for nodes too?

PR #797, opened 2024-12-31
Fix: nn-reassigned install selected when clicking node dot on map

Andrew-Dickinson:
FIxes a bug where when nodes with NNs that have been assigned since meshdb import are selected on the admin map, the old install that was re-assigned is selected in the edit pane instead of the node object

Also fixes other edge cases where member, building, and node objects are attached to an install in the NN Assigned or Closed status and are selected in the edit pane


WillNilges (approved):

WillNilges:
Automerging this because I need it for my map tomfoolery

PR #798, opened 2024-12-31
Add link from the NN Re-assigned install to the node that got its number

Andrew-Dickinson:
Screenshot 2024-12-31 at 00 44 10

Adds an "eye" icon to installs in the NN Reassigned status which links back to the node that got the install's number. Useful if you ever end up on the old install object instead of the node by mistake

Installs that are not "NN Reassigned" are unchanged:

Screenshot 2024-12-31 at 00 50 32


src/meshapi/widgets.py
```diff
@@ -111,3 +129,7 @@ def get_context(self, name: str, value: str, attrs: Optional[dict] = None) -> di
         context["auto_populate_source"] = self.source
         context["auto_populate_url"] = self.source
         return context
+
+
+# Down here to resolve circular imports
+from meshapi.admin.models.install import InstallAdminForm  # noqa: E402
```
WillNilges, line 135: Oooh that's a little cursed. Do you think there's another way to resolve the circular imports?

Andrew-Dickinson: I tried a few other things, but nothing works. The most common way to solve this is to use `import x; x.y` instead of `from x import y` but I think our liberal use of `from .x import *` statements in `__init__` is causing some issues with that

WillNilges: Woof. I guess we do this kind of thing in other places. Plus django admin panel is cursed software anyway.

WillNilges (approved):

PR #799, opened 2025-01-04
Update Admin Panel to work via iframes

WillNilges:
Refactors the way the map is incorporated into the admin panel. Instead of embedding it and doing a lot of tricks to update the DOM, it puts the admin panel and the map into their own iframes, and ties them together with a little JavaScript secret sauce.

- Adds ability to pick up where you left off. Your browser will remember the last page you visited and bring you back there if you go to `/admin`

Still a lot to do on this:
- [x] Capture forward, back, refresh, etc to make sure people don't lose their place
- [x] Resize + Hide the map
- [ ] Mobile
- [x] Refactoring
- [ ] Tests?


src/meshapi/admin/utils.py
```diff
@@ -18,7 +18,10 @@ def get_admin_url(model: Model, site_base_url: str) -> str:
         reverse(
             f"admin:{content_type.app_label}_{content_type.model}_change",
             args=(model.pk,),
-        ),
+        ).replace(
+            "panel/", ""
+        ),  # Strip "panel" out of admin URL. "Panel" views are meant to be shown
+        # inside the admin panel iframe, and likely the user does not want that
```
Andrew-Dickinson, line 24: Is there a better way to do this? `reverse()` is the canonical way to get a URL to something in Django, so if it's not returning the correct value that's not great

src/meshdb/templates/admin/base.html
```diff
@@ -113,9 +113,6 @@
             {% block footer %}<div id="footer"></div>{% endblock %}
           </div>
         </div>
-        {% block map_sidebar %}
-            {% include "admin/map_sidebar.html" %}
-        {% endblock %}
```
Andrew-Dickinson, line 118: Should we remove `map-wrapper` also? 

It's also worth diffing this file with the one it was copied from to see if we can just get rid of it altogether and take the upstream template

src/meshdb/templates/admin/map_sidebar.html
```diff
@@ -22,6 +22,6 @@
     <div id="map-scripts" style="display: none"></div>
     <script src="{% static '/admin/intercept.js' %}"></script>
     <script>let MAP_BASE_URL = "{% get_env_var 'ADMIN_MAP_BASE_URL' %}";</script>
-    <script src="{% static '/admin/map.js' %}"></script>
+    <!--<script src="{% static '/admin/map.js' %}"></script>-->
```
Andrew-Dickinson, line 25: Remove?

src/meshweb/static/admin/map.js
```diff
@@ -162,301 +183,162 @@ async function getNewSelectedNodes(){
...
+    admin_panel_iframe.removeEventListener("load", onAdminPanelLoad);
+}
 
-    let selectedNodeInt = parseInt(selectedNodes);
-    if (selectedNodeInt >= 1000000) {
-        /* Hack for APs to not break things. We unfortantely can't do a lot better than this without much pain*/
-        return;
+// Checks local storage for the last page the user navigated to, and directs them
+// there
+async function adminPanelRestoreLastVisited() {
```
Andrew-Dickinson, line 275: Do we really want this behavior?

src/meshapi/admin/utils.py
WillNilges: I don't like it either. That was my "get the fucking thing working" answer. I still need to root-cause it.

src/meshdb/templates/admin/map_sidebar.html
WillNilges: We can remove this whole file actually

src/meshweb/static/admin/map.js
WillNilges: Unclear. I used this to make the interface survive a reload without losing the user's place and realized as a side effect the admin interface now picks up where you left off. Could be kind of nifty. Maybe we just leave it and see if people complain?

src/meshdb/templates/admin/base.html
WillNilges: Diffed it. We include `admin-ext.css` and add this `flex` tag now.

```
░▒~/Code/nycmeshnet/meshdb/.venv▒░ [±wdn/map-iframe-2✘]
$ diff ./lib/python3.13/site-packages/django/contrib/admin/templates/admin/base.html ~/Code/nycmeshnet/meshdb/src/meshdb/templates/admin/base.html
15a16
> <link rel="stylesheet" href="{% static "admin/admin_ext.css" %}">
84c85
<     <div class="main" id="main">
---
>     <div class="main flex" id="main">
113a115
>
```

This also fixes the layout issues you called out in slack

src/meshapi/admin/utils.py
WillNilges: I think it's because that's how I've configured `admin.site.urls`. I might need to play with that some more.

WillNilges: One option: I can put the admin panel at a new URL (like "admin/panel" maybe) and do the redirect kinda the other way around. It'll make sense on my next push.

src/meshapi/admin/urls.py
```diff
@@ -1,17 +1,19 @@
...
     AdminPasswordResetView,
 )
+from meshdb.views import admin_iframe_view
 
 urlpatterns = [
     path("password_reset/", AdminPasswordResetView.as_view(), name="admin_password_reset"),
     path("password_reset/done/", AdminPasswordResetDoneView.as_view(), name="password_reset_done"),
     path("password_reset/<uidb64>/<token>/", AdminPasswordResetConfirmView.as_view(), name="password_reset_confirm"),
     path("password_reset/done/", AdminPasswordResetCompleteView.as_view(), name="password_reset_complete"),
+    re_path(r"^panel/.*$", admin_iframe_view),  # Match any other /admin/* URL
```
Andrew-Dickinson, line 17: Does this still need to be so broad? With the JS solution, why does it still need a regex? Also perhaps we should be a bit more descriptive and call it `"iframe_wrapper/"` or something

src/meshdb/templates/admin/password_reset/password_reset_complete.html
```diff
@@ -1,3 +0,0 @@
-{% extends "registration/password_reset_complete.html" %}
-
-{% block map_sidebar %}{% endblock %}
```
Andrew-Dickinson, line 3: Hmm did you confirm these pages still work correctly and look good? I think the reason we had this before is that the map didn't look right on them. Also does this mean we show the map on the login page? I don't have a particular problem with that but it feels a bit odd

src/meshweb/static/admin/map.js
Andrew-Dickinson: What do you mean "survive a reload"? Doesn't the URL keep track of the page they're on?

Andrew-Dickinson: I don't love keeping state separate from the browser's own state, I feel like that exposes us to potential race conditions that jerk the user to an unexpected place

src/meshdb/templates/admin/password_reset/password_reset_complete.html
WillNilges: Whoops, thanks. I put those back. Login page doesn't have the map on it anymore, but there's currently a bug where the logout page does. I have a hamfisted idea to fix that.

src/meshweb/static/admin/map.js
WillNilges: > survive a reload

There was a point when reloading would send you back to the home, because I didn't do the URL stuff yet. Now that I have, we don't need this anymore.

src/meshapi/admin/urls.py
WillNilges: Evidently not, actually.

src/meshdb/templates/admin/password_reset/password_reset_complete.html
WillNilges: fixed tm

src/meshapi/admin/urls.py
```diff
@@ -7,11 +7,13 @@
...
     AdminPasswordResetView,
 )
+from meshdb.views import admin_iframe_view
 
 urlpatterns = [
     path("password_reset/", AdminPasswordResetView.as_view(), name="admin_password_reset"),
     path("password_reset/done/", AdminPasswordResetDoneView.as_view(), name="password_reset_done"),
     path("password_reset/<uidb64>/<token>/", AdminPasswordResetConfirmView.as_view(), name="password_reset_confirm"),
     path("password_reset/done/", AdminPasswordResetCompleteView.as_view(), name="password_reset_complete"),
+    path("iframe_wrapper/", admin_iframe_view),  # Match any other /admin/* URL
```
Andrew-Dickinson, line 17:
```suggestion
    path("iframe_wrapper/", admin_iframe_view),
```

src/meshapi/tests/test_admin_panel.py
```diff
@@ -0,0 +1,107 @@
...
+        soup = BeautifulSoup(decoded_panel, "html.parser")
+        iframe = soup.find(id="admin_panel_iframe")
+        iframe_src = iframe.attrs["src"]
+        self.assertEqual("/admin/", iframe_src)
+        iframe_response = self.c.get(iframe_src)
+        self.assertEqual(code, iframe_response.status_code, f"Could not view {route} in the admin panel.")
+
+    # TODO (wdn): Add more tests checking if navigating to xyz page works
+    # Unfortunately, because that is a lot of javascript, it's tricky to test.
+    # It may be possible to run selenium integration tests or something to validate
```
Andrew-Dickinson, line 106: I'd support selenium-based tests for integration testing deployed copies of the application (dev, gamma, etc) but probably we don't want it in our unit testing pipeline, those things are slow enough already

src/meshweb/static/admin/iframe_check.js
```diff
@@ -0,0 +1,62 @@
+async function checkIframed() {
+  // Check if we're in an iframe
+  const inIframe = window.self !== window.top;
+  
+  // If we are, do nothing
+  if (inIframe) {
```
Andrew-Dickinson, line 6: FWIW I think you should handle mobile by just adding an `|| isMobile()` here, so they don't deal with any of the iframe stuff and have just a pure HTML experience with no map

Andrew-Dickinson:
Doing some local testing and noticed a few issues:

- After clicking "logout" in the header, the map is still shown until refreshing the page
- Clicking "View Site" in the header results in a weird page where the map is shown next to the landing page
- LOSes and APs don't seem to show on the map when you select them in the left panel

Overall it feels much snappier though, great job on the performance improvements

WillNilges:
> After clicking "logout" in the header, the map is still shown until refreshing the page

That's odd. My last commit should have fixed that. Did you pull the latest?

Andrew-Dickinson (approved):
I'm not looking super closely, but I don't want to block you any more than I already have. You've iterated on this a ton and I trust your judgement about the readiness of this for release

src/meshweb/static/admin/iframe_check.js
```diff
@@ -0,0 +1,55 @@
...
+
+    mapWrapper.parentNode.insertBefore(
+        errorNotice,
+        mapWrapper
+    );
+    return;
+  }
+  
+  // FIXME (wdn): This might not totally clear the state after all, since the variables
+  // from /admin are still hanging around
```
WillNilges, line 49:
```suggestion
```

src/meshdb/views.py
```diff
@@ -0,0 +1,8 @@
+from django.contrib.admin.views.decorators import staff_member_required
+from django.http import HttpRequest, HttpResponse
+from django.shortcuts import render
+
+
+@staff_member_required
+def admin_iframe_view(request: HttpRequest) -> HttpResponse:
```
WillNilges, line 7: Might want to add a test for this authentication/redirect

src/meshweb/static/admin/map.js
```diff
@@ -1,28 +1,52 @@
-
+const ADMIN_PANEL_HOME = "/admin/"
+const MESHDB_LAST_PAGE_VISITED = "MESHDB_LAST_PAGE_VISITED"
+const admin_panel_iframe = document.getElementById("admin_panel_iframe");
```
WillNilges, line 3: camelCase this

src/meshweb/static/admin/map.js
```diff
@@ -162,301 +186,145 @@ async function getNewSelectedNodes(){
...
+        }
     }
 
-    if (target.className === "capture-exclude") return true;
-    if (!url.pathname.startsWith("/admin/")) return true;
-    if (url.pathname.startsWith("/admin/login")) return true;
-    if (url.pathname.startsWith("/admin/logout")) return true;
-    if (url.pathname.endsWith("/export/") && isForm) return true;
-    if (url.host !== location.host) return true;
+    console.log(`Admin Panel src set to: ${admin_panel_iframe.src}`);
```
WillNilges, line 220: Might want to gate all these log lines with a debug statement or something for cleanliness.

src/meshweb/static/admin/map.js
```diff
@@ -496,18 +375,22 @@ function allowMapResize() {
...
 function hideMapIfAppropriate() {
     const isMobile = mobileCheck();
 
     const mapDisabled = localStorage.getItem("MESHDB_MAP_DISABLED") === "true" || isMobile;
     if (mapDisabled) {
-        document.getElementById('map').classList.add("hidden");
-        document.getElementById('main').classList.remove("flex");
+        document.getElementById('map_panel_div').classList.add("hidden");
+        document.getElementById('map_controls').classList.add("hidden");
+        //document.getElementById('main').classList.remove("flex");
```
WillNilges, line 393:
```suggestion
```

PR #802, opened 2025-01-07
Fix: Installs whose number has been re-used as an NN shouldn't be able to get an NN of their own

Andrew-Dickinson:
Fixes an issue where after an old install number is recycled as a network number for another node, that old install number is still able to get an NN of its own. This is needlessly confusing and creates extra edge cases when trying to disambiguate a number that is presented into a NN or Install Number. It's not a big deal to ask these folks to just fill out the join form again, so we disallow this from happening


src/meshapi/tests/test_nn.py
```diff
@@ -429,25 +429,26 @@ def test_nn_valid_low_install_number_used_nn(self):
...
-        resp_nn = json.loads(response.content.decode("utf-8"))["network_number"]
-        expected_nn = 101
         self.assertEqual(
-            expected_nn,
-            resp_nn,
-            f"nn incorrect for test_nn_valid_install_number. Should be {expected_nn}, but got {resp_nn}",
+            json.loads(response.content.decode("utf-8")),
+            {
+                "detail": "Invalid install status for NN Assignment. "
+                "Re-submit the join form to create a new install number"
```
WillNilges, line 443: Just a thought: Should we have some one-click way to re-submit a join form for an install whose NN has been re-assigned?

src/meshapi/views/forms.py
```diff
@@ -516,6 +517,18 @@ def network_number_assignment(request: Request) -> Response:
...
+        # Don't allow old recycled or closed installs to have a network number assigned to them,
+        # since this would be very confusing. Installs in this status should re-submit the join
+        # form and try again with a new install number
+        if nn_install.status in [Install.InstallStatus.CLOSED, Install.InstallStatus.NN_REASSIGNED]:
+            return Response(
+                {
+                    "detail": "Invalid install status for NN Assignment. "
+                    "Re-submit the join form to create a new install number"
+                },
+                status=status.HTTP_409_CONFLICT,
```
WillNilges, line 529: This is a good code

src/meshapi/views/forms.py
```diff
@@ -516,6 +517,18 @@ def network_number_assignment(request: Request) -> Response:
 
     # First, we try to identify a Node object for this install if it doesn't already have one
     if not nn_install.node:
+        # Don't allow old recycled or closed installs to have a network number assigned to them,
+        # since this would be very confusing. Installs in this status should re-submit the join
+        # form and try again with a new install number
+        if nn_install.status in [Install.InstallStatus.CLOSED, Install.InstallStatus.NN_REASSIGNED]:
```
WillNilges, line 523: I guess this is just how it has to be for reassigned NNs, but what do we think about deduping closed installs? Maybe this is an edge case not worth thinking about, but if it takes someone 3 tries to get installed after filling out the join form, they're gonna have three install numbers which would clutter the UI.

Then again, we can just... re-open the request or something.

WillNilges (approved):

src/meshapi/tests/test_nn.py
Andrew-Dickinson: Seems like such an edge case, copy pasting from another browser tab is really not that bad

src/meshapi/views/forms.py
Andrew-Dickinson: I'm not sure I follow what you're saying about closed installs. Can you elaborate on the case you're worried about?

WillNilges: What happens if someone submits the join form, then their request gets closed, then they submit the join form again? Does it re-open the old request or make a new one with the same information?

Andrew-Dickinson: If the request is closed, they will get a new install number

Andrew-Dickinson: I think this is a data safety issue. Unless we can prove that in all cases it would be safe to re-open a closed request, we have to error on the side of caution and issue a new install number

PR #803, opened 2025-01-07
Set max network number to 8000 instead of 8192

Andrew-Dickinson:
The docs and configen say the max number is `8000` and we should be consistent with that

Also enforces minimum and maximum NN at node creation time


WillNilges (commented):
If MeshDB can support it, why not update docs/configgen/etc and take advantage of the extra NNs?

src/meshapi/tests/test_map_endpoints.py
WillNilges, line 1: Why were all these higher than 8192 in the first place? Were we not enforcing them before?

WillNilges (approved):

WillNilges:
Talked about this in slack. Things will get kind of messy, and it's not worth trying to get an extra 192 addresses in this PR.

src/meshapi/tests/test_map_endpoints.py
Andrew-Dickinson: Good question. The PR is too old and I don't remember lol

PR #804, opened 2025-01-07
Readiness + Startup probes

james-otten:
The liveness probe was not working because the element was duplicated with the second being empty. Re-enable it, make it work, and add readiness + startup probes to allow for slow startups, while pulling pods from the service and eventually restarting them if they are unresponsive.

Let's let this soak in dev for a while.

### DIFF
--- a/infra/helm/meshdb/templates/meshweb.yaml
+++ b/infra/helm/meshdb/templates/meshweb.yaml
@@ -47,18 +47,30 @@ spec:
               mountPath: /opt/meshdb/static
           {{ if eq .Values.meshweb.liveness_probe "true" }}
           livenessProbe:
-           exec:
-             command:
-               - curl
-               - http://127.0.0.1:{{ .Values.meshweb.port }}/api/v1
-           periodSeconds: 3
-           initialDelaySeconds: 4
-           timeoutSeconds: 3
-           {{ end }}
-          livenessProbe:
-            {{- toYaml .Values.livenessProbe | nindent 12 }}
+            httpGet:
+              path: /api/v1
+              port: {{ .Values.meshweb.port }}
+              scheme: http
+              httpHeaders:
+                - name: host
+                  value: 127.0.0.1
+            periodSeconds: 3
+            initialDelaySeconds: 4
+            timeoutSeconds: 3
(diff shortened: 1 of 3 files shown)


Andrew-Dickinson (approved):

infra/helm/meshdb/templates/meshweb.yaml
```diff
@@ -47,18 +47,38 @@ spec:
...
-           initialDelaySeconds: 4
-           timeoutSeconds: 3
-           {{ end }}
-          livenessProbe:
-            {{- toYaml .Values.livenessProbe | nindent 12 }}
+            exec:
+              command:
+                - bash
+                - -c
+                - 'curl http://127.0.0.1:{{ .Values.meshweb.port }}/api/v1/ -H "Host: db.nycmesh.net" -s | grep meshin'
```
WillNilges, line 54: The grep is fine but knowing that it's `we're meshin'` confused my brain for a sec lol

infra/helm/meshdb/templates/meshweb.yaml
```diff
@@ -47,18 +47,38 @@ spec:
...
+            periodSeconds: 3
+            initialDelaySeconds: 20
+            timeoutSeconds: 3
+            failureThreshold: 20
+          {{ end }}
```
WillNilges, line 81: I don't think it's worth having redundant probes. We should consider what kind of probes to use in place of these. Maybe we should just check if we get anything back from nginx at all as the liveness probe, then make the readiness probe check `we're meshin'`?

Not sure about startupProbe

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

WillNilges: Actually, I think I might be wrong to suggest that.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

> A common pattern for liveness probes is to use the same low-cost HTTP endpoint as for readiness probes, but with a higher failureThreshold. This ensures that the pod is observed as not-ready for some period of time before it is hard killed.

WillNilges (approved):

PR #805, opened 2025-01-07
Add number disambiguation endpoint

### LINKED ISSUE #801: Expose an API for identifying ambiguous numbers
Create an API that uses MeshDB data to take an integer parameter and identify if it corresponds to a network number or install number. e.g.
```
73 -> Neither
328 -> Install Number
238 -> Network Number
714 -> Network and Install Number (different place)
5916 -> Network and Install Number (same place)
12341 -> Install Number
```

Perhaps we should also give out the UUIDs of each of these objects?

Andrew-Dickinson:
Adds a new API endpoint which aids in network/install number disambiguation. Given an ambiguous number, checks if it matches the NN for an existing node, and returns that node. If not, checks the given number as an install number and returns the attached node (if it exists).

Also returns fields to "show its work", which let you see what objects it found matching the requested number

Closes #801


src/meshapi/views/helpers.py
```diff
@@ -0,0 +1,128 @@
...
+install_serializer_field = InstallNestedRefSerializer()
+node_serializer_field = NestedKeyObjectRelatedField(
+    queryset=Node.objects.all(),
+    additional_keys=("network_number",),
+)
+
+
+# TODO: Fix docs, maybe with custom install serializer?
+class DisambiguateInstallOrNetworkNumber(APIView):
+    permission_classes = [HasDisambiguateNumberPermission]
```
WillNilges, line 24: What do you think about just letting anyone with read-only role or better use this instead of making a permission? Is there a service that needs least privilege?

src/meshapi/views/helpers.py
```diff
@@ -0,0 +1,128 @@
...
+
+        output = {
+            "resolved_node": node_serializer_field.to_representation(resolved_node) if resolved_node else None,
+            "supporting_data": {
+                "exact_match_recycled_install": (
+                    install_serializer_field.to_representation(install_object)
+                    if install_object and install_object.status == Install.InstallStatus.NN_REASSIGNED
+                    else None
+                ),
+                "exact_match_nonrecycled_install": (
```
WillNilges, line 119: Nit, I think it would be clearer to have `exact_match_install`:
```suggestion
                "exact_match_install": (
```

WillNilges (commented):
A couple initial comments on this. Logic of the actual endpoint looks sound.

src/meshapi/views/helpers.py
Andrew-Dickinson: I actually have a project in the works literally right now that needs just this permission and nothing else. Also @james-otten had a use for this too, with his device provisioning kiosk idea

So I'd guess based on that it's not uncommon

james-otten (approved):

PR #806, opened 2025-01-10
Warn the user if they forgot to set install/abandon date

Andrew-Dickinson:
Adds a bit of javascript to the Install and Node admin forms which reminds the admin to set the install/abandon date as appropriate if they forget to set it. It's just a reminder but should hopefully encourage folks to actually set these fields

https://github.com/user-attachments/assets/8ba81c0a-1ff8-4276-9d86-24c3cff500cb

### DIFF
--- a/src/meshapi/static/widgets/warn_about_date.js
+++ b/src/meshapi/static/widgets/warn_about_date.js
@@ -32,41 +32,28 @@ $(document).ready(function($) {
     }
 
     function showWarning(dateField, statusField, saveAction, dateType){
-      const saveCurrentDateButton = document.createElement("a");
-        saveCurrentDateButton.className = "button";
-        saveCurrentDateButton.onclick = function () {
+        const errorNotice = $("#dateMissingWarningTemplate").clone();
+        errorNotice.prop("id", "dateMissingWarning");
+
+        errorNotice.find("#saveCurrentDateButton").on("click", () => {
             dateField.val(getCurrentISODate())
             submitAdminForm(saveAction);
             return false;
-        }
-        saveCurrentDateButton.style = "padding: 10px 15px; display: inline-block; text-decoration: none;";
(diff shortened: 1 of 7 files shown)


src/meshapi/static/widgets/warn_about_date.js
```diff
@@ -0,0 +1,111 @@
...
+        `;
+        errorNotice.appendChild(saveCurrentDateButton);
+        errorNotice.appendChild(saveNoDateButton);
+
+        saveButtons.each((i, button) => {
+            $(button).addClass("disabled-button");
+        })
+
+        $(errorNotice).insertBefore($('.submit-row').first())
+    }
```
WillNilges, line 70: Was it really easier to do this than to create another html template? I guess if it works, but this isn't terribly readable.

src/meshapi/static/widgets/warn_about_date.js
```diff
@@ -0,0 +1,111 @@
...
+        } else if (["Closed", "Inactive"].indexOf(status) !== -1){
+            if (installDate && !abandonDate) {
+                hideWarning();
+                showWarning($("#id_abandon_date"), $('#id_status'), e.target.name, "Abandon")
+
+                return false;
+            }
+        }
+
+        return true;
```
WillNilges, line 109: Dumb question: The act of returning false/true controls whether or not the form actually gets submitted?

src/meshapi/static/widgets/warn_about_date.js
Andrew-Dickinson: I’ll try to implement a template based solution tonight and see how it looks. The nice thing about this approach is that it’s modular and could be used alongside other form widgets

WillNilges: I suppose so, and I guess if it works it works...

WillNilges (approved):

PR #810, opened 2025-01-20
fix crash in scramble script by adding try/except guard on node saving

WillNilges:
Title. Ran into this while getting a new dev online

### DIFF
--- a/src/meshapi/management/commands/scramble_members.py
+++ b/src/meshapi/management/commands/scramble_members.py
@@ -117,11 +117,14 @@ class Command(BaseCommand):
         logging.info("Scrambling nodes...")
         nodes = Node.objects.all()
         for node in nodes:
-            node.notes = fake.text()
-            _, node.install_date, node.abandon_date = self.fuzz_dates(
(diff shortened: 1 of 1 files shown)


src/meshapi/management/commands/scramble_members.py
```diff
@@ -37,7 +36,9 @@ def add_arguments(self, parser: ArgumentParser) -> None:
             help="Skip scrambling installs",
         )
 
-    @transaction.atomic
+    # This transaction is not atomic so that even if there's some error with
+    # the scrambled data, we can scramble it partially. I'd rather break someone's
+    # dev env than give someone PII accidentally
```
Andrew-Dickinson, line 41: Doesn't removing the atomic block make it more likely that we will disclose PII? Since a single row could fail silently?

WillNilges: whoops, I forgot to push again. I have another commit that sets up atomic for each individual transaction.

Andrew-Dickinson (approved):

PR #817, opened 2025-01-29
add celery task to dedupe history

WillNilges:
⚠️ Still can't get this working reliably in prod. Do not merge.


WillNilges:
This is probably not a great idea, considering it corrupted our database last time. I'm just going to write some docs on the wiki to explain how I un-bloated the database.

PR #819, opened 2025-01-30
Don't call save on devices, links, or LOSes unnecessarily

### LINKED ISSUE #790: History entries are created for most models when not making any changes
When not making any changes and clicking "Save", no entries should be created in the model history, but for most models this currently happens due to a variety of issues like notes changing from `None -> ""` or loosing precision in datetimes

WillNilges:
Closes #790 

This prevents the UISP import script from calling save on our objects unnecessarily, which should stop the history tables from ballooning like they do.

See https://django-simple-history.readthedocs.io/en/3.8.0/utils.html#clean-duplicate-history for more info

- [x] Write Tests?

### DIFF
--- a/src/meshapi/util/uisp_import/sync_handlers.py
+++ b/src/meshapi/util/uisp_import/sync_handlers.py
@@ -375,11 +375,16 @@ def sync_link_table_into_los_objects() -> None:
                         changed_los = True
 
                     # Keep the LOS analysis date accurate for all that come from existing links
-                    if existing_los.source == LOS.LOSSource.EXISTING_LINK and link.last_functioning_date_estimate:
+                    if (
+                        existing_los.source == LOS.LOSSource.EXISTING_LINK
+                        and link.last_functioning_date_estimate
+                        and existing_los.analysis_date != link.last_functioning_date_estimate
+                    ):
                         existing_los.analysis_date = link.last_functioning_date_estimate
                         changed_los = True
 
                     if changed_los:
+                        print("changed los")
                         existing_los.save()
                 continue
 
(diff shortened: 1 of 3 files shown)


src/meshapi/util/uisp_import/sync_handlers.py
```diff
@@ -129,6 +132,8 @@ def import_and_sync_uisp_devices(uisp_devices: List[UISPDevice]) -> None:
                 uisp_device["identification"]["model"], DEFAULT_SECTOR_WIDTH
             )
 
+            # Only when we're sure the sector doesn't exist do we save it
+            # XXX (wdn): is this the only place we try to save sectors in here?
```
WillNilges, line 136: AFAICT yes

WillNilges:
Confirmed this test fails on main and works on my branch.

src/meshapi/util/uisp_import/update_objects.py
```diff
@@ -104,7 +106,7 @@ def update_link_from_uisp_data(
 
     if uisp_link_id != existing_link.uisp_id:
         existing_link.uisp_id = uisp_link_id
-        logging.info(
+        change_messages.append(
```
Andrew-Dickinson, line 109: You don’t want this. It’s super spammy. Like 10 messages an hour spammy

src/meshapi/util/uisp_import/sync_handlers.py
```diff
@@ -241,6 +246,8 @@ def import_and_sync_uisp_links(uisp_links: List[UISPDataLink]) -> None:
                     message=f"Possible duplicate objects detected, links share the same UISP ID ({uisp_uuid})",
                 )
 
+            # XXX (wdn): Do we want to make a history record in this case? I _think_ so?
+            # maybe i dont understand this well enough
```
Andrew-Dickinson, line 250: Yeah I think we do want a record for this, since it's changing something we care about. It might be noisy, but that's UISP's fault :(

Andrew-Dickinson (approved):
I think we're gonna be unhappy with how spammy the UISP id rotation slack messages are, but we can try it

src/meshapi/util/uisp_import/update_objects.py
WillNilges: I'll set dirty bit on this, then.

Andrew-Dickinson (approved):

PR #836, opened 2025-02-12
Join Record Replay improvements

WillNilges:
- Small refactoring of the replay system
- some better logs
- adds a simple web interface for viewing join records
- inject join record env vars into container

### DIFF
--- a/.github/workflows/deploy-to-k8s.yaml
+++ b/.github/workflows/deploy-to-k8s.yaml
@@ -59,6 +59,8 @@ jobs:
         --set email.smtp_password="${{ secrets.SMTP_PASSWORD }}" \
         --set meshweb.backup_s3_base_folder="${{ vars.BACKUP_S3_BASE_FOLDER }}" \
         --set meshweb.backup_s3_bucket_name="${{ secrets.BACKUP_S3_BUCKET_NAME }}" \
+        --set meshweb.join_record_bucket_name="${{ secrets.JOIN_RECORD_BUCKET_NAME }}" \
+        --set meshweb.join_record_prefix="${{ secrets.JOIN_RECORD_PREFIX }}" \
         --set meshweb.django_secret_key="${{ secrets.DJANGO_SECRET_KEY }}" \
         --set meshweb.nn_assign_psk="${{ secrets.NN_ASSIGN_PSK }}" \
         --set meshweb.query_psk="${{ secrets.QUERY_PSK }}" \
--- a/src/meshapi/management/commands/replay_join_records.py
+++ b/src/meshapi/management/commands/replay_join_records.py
@@ -42,17 +42,26 @@ class Command(BaseCommand):
         )
 
     def handle(self, *args: Any, **options: Any) -> None:
-        logging.info("Fetching Join Records...")
+        # Default to getting join records from 1 week ago unless otherwise specified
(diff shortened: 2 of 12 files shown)


src/meshapi/management/commands/replay_join_records.py
```diff
@@ -42,18 +42,27 @@ def add_arguments(self, parser: ArgumentParser) -> None:
         )
 
     def handle(self, *args: Any, **options: Any) -> None:
-        logging.info("Fetching Join Records...")
+        # Default to getting join records from 1 week ago unless otherwise specified
+        since = options["since"] or self.past_week()
```
james-otten, line 46: Does `options["since"]` need to be converted to `datetime`?

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -59,6 +59,8 @@ jobs:
         --set email.smtp_password="${{ secrets.SMTP_PASSWORD }}" \
         --set meshweb.backup_s3_base_folder="${{ vars.BACKUP_S3_BASE_FOLDER }}" \
         --set meshweb.backup_s3_bucket_name="${{ secrets.BACKUP_S3_BUCKET_NAME }}" \
+        --set meshweb.join_record_bucket_name="${{ secrets.JOIN_RECORD_BUCKET_NAME }}" \
```
james-otten, line 62: Add both to dev3 + gamma envs in UI. `| b64enc` will have a bad day

src/meshweb/views/join_record_viewer.py
```diff
@@ -0,0 +1,43 @@
...
+from botocore.exceptions import ClientError
+from django.contrib.admin.views.decorators import staff_member_required
+from django.http import HttpRequest, HttpResponse
+from django.template import loader
+
+from meshapi.management.commands import replay_join_records
+from meshapi.util.join_records import JoinRecordProcessor
+
+
+@staff_member_required
```
james-otten, line 13: Super?

james-otten (approved):

src/meshweb/views/join_record_viewer.py
WillNilges: Good catch. That was not my intention, but on further thought, this is just a viewer; I think it's valid for staff members (olvier, Matthew, etc) to be able to _view_ join records that need replaying.

src/meshapi/management/commands/replay_join_records.py
WillNilges: There is a [lambda](https://github.com/nycmeshnet/meshdb/pull/836/files#diff-aed32b816222e67d35fdcf17ed956a53bb18f387e96dd5b6cb22c6d3daa91d53R34) on the command line argument that converts it to datetime.

.github/workflows/deploy-to-k8s.yaml
WillNilges: Added. Side note: the dev3 prefix is called `dev3-join-form-submissions`, but we should delete the prefix and make a new one.

james-otten: Thanks

src/meshweb/views/join_record_viewer.py
james-otten: Sounds good for view.

PR #838, opened 2025-02-21
Don't drop tables on import and specify database

WillNilges:


scripts/import_spreadsheet_dump.sh
```diff
@@ -40,22 +40,8 @@ do
...
-#    docker exec -i meshdb_postgres_1 pg_dump -U meshdb --table="$table_name" > "$table_name.sql"
-#done
-
 num_tables=${#tables[@]}
 
-# Yeet
-# XXX (willnilges): Would it be better to use manage.py?
-for ((i = num_tables - 1; i >= 0; i--));
-do
-	$DOCKER_PG_COMMAND -c "DROP TABLE IF EXISTS ${tables[i]} CASCADE"
```
Andrew-Dickinson, line 56: So looking at it now it's coming back to me why this might be needed. I think this should acutally be `DELETE FROM` for the type of export I generated for Sam :(

The import script ones were exported directly from `psql` with a command that might have been lost to time. These ones were generated with datagrip, and apparently they don't include the table definitions, just the insert statements. We should tune this script to whatever mechanized export process we want going forward (since the import script is no more)

WillNilges: Oh interesting, I didn't realize you had used a separate import mechanism. I agree that this process needs some work, but we should not rely on datagrip for this. My favorite thing to do would probably be to grab a backup from the S3 bucket, and write a script (perhaps a manage.py script) that could do the import/scramble/export in one shot.

WillNilges:
This will require some noodling. Closing for now.

WillNilges:
https://github.com/nycmeshnet/meshdb/issues/845

PR #844, opened 2025-03-01
Set up app-level tracing

WillNilges:
This installs and configures ddtrace for MeshDB, so we should get good django tracing without some of the overhead of single step instrumentation.

I am not sure exactly _how_ much overhead this will have, so I'd say we should disable single-step and try it out for a while.


infra/helm/meshdb/charts/celery/templates/deployment.yaml
```diff
@@ -39,7 +39,10 @@ spec:
           resources:
             {{- toYaml $.Values.resources | nindent 12 }}
           command: {{ toJson .command }}
-          # TODO (willnilges): Fix this in a later PR
+          env:
+          - name: DD_SERVICE
+            value: celery 
+          # TODO (willnilges): Fix this in a later PR. Computers
```
WillNilges, line 45:
```suggestion
          # TODO (willnilges): Fix this in a later PR.
```

WillNilges:
I can now see celery ticks, heartbeat, etc in Datadog. I'd like to confirm that traces look the same or better before I merge.

Also, what is important for us to trace:
- [x] MeshDB
- [ ] Meshforms (will handle separately)
- [x] Celery™
- [ ] Postgres? Not as important. We should see how it looks and go from there.

Celery looks good.
image

WillNilges:
This is what we're getting now with SSI. It's good, but there's quite a bit of what I'd call cruft in here.

image

WillNilges:
Possibly of note, service names will change:
- `meshdb-meshweb` -> `meshdb`
- `meshdb-celery` -> `celery`

I could update this so that doesn't happen, if folks feel strongly about it.

This change will likely improve tagging on things like our logs.

Andrew-Dickinson (approved):

Andrew-Dickinson:
> Possibly of note, service names will change:
> - `meshdb-meshweb` -> `meshdb`
> - `meshdb-celery` -> `celery`

Huge plus. Meshweb is super confusing

PR #846, opened 2025-03-03
use rolling updates for meshdb

WillNilges:
This should prevent more than one meshdb pod from coming up/going down at a time when rolling out.

We might want to do something similar for postgres, but I am not sure if that is safe.


james-otten (approved):
Let's give it a bit for the dust to settle

PR #858, opened 2025-03-09
Add additional_members field

WillNilges:
TODO:
- [x] Squash migrations
- [x] Write some tests?
    - [x] Ensure that the fields are properly queryable (when you query for an install, it shows up, when you query a member, all of their nodes show up)


WillNilges:
image

Andrew-Dickinson (approved):

src/meshapi/models/install.py
```diff
@@ -105,6 +105,12 @@ class InstallStatus(models.TextChoices):
         related_name="installs",
         help_text="The member this install is associated with",
     )
+    additional_members = models.ManyToManyField(
+        Member,
+        blank=True,
+        related_name="additional_installs",
+        help_text="Any additional members associated with this install",
```
Andrew-Dickinson, line 112:
```suggestion
        help_text="Any additional members associated with this install. E.g. roommates, parents, caretakers etc. Anyone that might contact us on behalf of this install belongs here",
```

WillNilges:
Got feedback on the widget. Ideally, we probably want an inline so that we can see details about the member, and add new ones.

PR #862, opened 2025-03-16
Fix

james-otten:
Fix extra white space issue, leading to the helm upgrade command being incorrect, and this thing not being able to deploy.

Edit: Also, make text editors catch this.


Andrew-Dickinson (approved):
😭

PR #870, opened 2025-03-21
Add on-demand UISP import by NN

WillNilges:
image


src/meshapi/util/uisp_import/sync_handlers.py
```diff
@@ -63,6 +63,9 @@ def import_and_sync_uisp_devices(uisp_devices: List[UISPDevice]) -> None:
         # represent the other side of the link
         uisp_network_number = int(network_number_matches[0])
 
+        if target_network_number and uisp_network_number != target_network_number:
+            pass
```
WillNilges, line 67:
```suggestion
            continue
```

Andrew-Dickinson:
You could also just skip them and let it happen on the next hourly cycle. They’re a lot less important

WillNilges:
I made him a little prettier.
image

src/meshapi/urls.py
```diff
@@ -39,6 +39,7 @@
         views.DisambiguateInstallOrNetworkNumber.as_view(),
         name="meshapi-v1-disambiguate-number",
     ),
+    path("uisp-import/nn/<network_number>/", views.uisp_import_for_nn, name="uisp-import-by-nn"),
```
Andrew-Dickinson, line 42: Should there be a type here:

Andrew-Dickinson, line 42: Maybe there should be a type here: `<int:network_number>`? I think that makes it 404 automatically if they give us letters

src/meshapi/util/uisp_import/sync_handlers.py
```diff
@@ -228,6 +234,14 @@ def import_and_sync_uisp_links(uisp_links: List[UISPDataLink]) -> None:
             )
             continue
 
+        # If we're crawling a specific NN and neither of the NNs are
```
Andrew-Dickinson, line 237: nit: incomplete comment

src/meshapi/views/uisp_import.py
```diff
@@ -0,0 +1,74 @@
...
+            response=OpenApiTypes.STR,
+        ),
+    },
+)
+@api_view(["POST"])
+@permission_classes([IsAdminUser])
+def uisp_import_for_nn(request: Request, network_number: int) -> Response:
+    logging.info(f"Received uisp import request for NN{network_number}")
+    if not network_number:
+        status = 400
```
Andrew-Dickinson, line 42: Maybe this should be 404?

src/meshapi/views/uisp_import.py
```diff
@@ -0,0 +1,74 @@
...
+        return Response({"detail": m}, status=status)
+
+    try:
+        target_nn = int(network_number)  # Because I apparently can't trust nobody
+        # Must be in valid range.
+        # NETWORK_NUMBER_MIN is set to something meshdb can assign new stuff, but we have lower NNs
+        # hence, I hardcode 1
+        FIRST_NN = 1
+        if target_nn < FIRST_NN or NETWORK_NUMBER_MAX < target_nn:
+            status = 400
```
Andrew-Dickinson, line 54: Same as above, 404?

src/meshapi/views/uisp_import.py
```diff
@@ -0,0 +1,74 @@
...
+            m = "Network number is not in valid range."
+            logging.error(m)
+            return Response({"detail": m}, status=status)
+    except ValueError:
+        status = 400
+        m = f"Network Number must be an integer between {NETWORK_NUMBER_MIN} and {NETWORK_NUMBER_MAX}."
+        return Response({"detail": m}, status=status)
```
Andrew-Dickinson, line 61: IMO we could simplify this to just change it to try to fetch a `Node` object with that network number. It requires NN assignment to have happened already, but like, that's a feature not a bug

I think maybe the UISP import logic requires the node to exist anyway, so it's not like we're doing them any favors here. If there are devices associated with the NN they enter but no `Node` object, I think the UISP import is essentially a no-op

Andrew-Dickinson (approved):

src/meshapi/views/uisp_import.py
WillNilges:
```suggestion
        status = 404
```

src/meshapi/views/uisp_import.py
WillNilges:
```suggestion
            status = 404
```

src/meshapi/urls.py
WillNilges:
```suggestion
    path("uisp-import/nn/<int:network_number>/", views.uisp_import_for_nn, name="uisp-import-by-nn"),
```

src/meshapi/util/uisp_import/sync_handlers.py
WillNilges:
```suggestion
        # If we're importing from a specific NN and neither of the NNs are the one we expect, then ignore this link and move on.
```

WillNilges:
Codecov, I have bent over for you 3 times. Enough is enough.

PR #872, opened 2025-03-23
draft PR

vvedenskaya:


src/meshapi/pelias.py
```diff
@@ -4,10 +4,11 @@
 
 import inflect
 import requests
+import environment
 
 from meshapi.util.constants import DEFAULT_EXTERNAL_API_TIMEOUT_SECONDS
 
-PELIAS_ADDRESS_PARSER_URL = os.environ.get("PELIAS_ADDRESS_PARSER_URL", "http://localhost:6800/parser/parse")
+PELIAS_ADDRESS_PARSER_URL = environm
```
WillNilges, line 11: You should be able to do something like this:
```suggestion
from environment import PELIAS_ADDRESS_PARSER_URL
```

src/meshdb/celery.py
```diff
@@ -1,11 +1,12 @@
 import os
 from pathlib import Path
-
+from django.conf import settings
 from celery import Celery, bootsteps
 from celery.apps.worker import Worker
 from celery.signals import beat_init, worker_ready, worker_shutdown
 from datadog import initialize, statsd
 from dotenv import load_dotenv
+import environment
```
WillNilges, line 9: Consider only importing what you need.
```suggestion
from environment import CELERY_BROKER
```

src/meshdb/environment.py
WillNilges, line 1: Radical! Thanks for including where the env vars are from, too 😄

WillNilges (commented):
This is definitely on the right track, thanks so much for working on this!!!

__init__.py
```diff
@@ -0,0 +1,2 @@
+def environment():
```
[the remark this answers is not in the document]
vvedenskaya: deleted, my bad

src/meshdb/celery.py
```diff
@@ -64,7 +65,7 @@ def beat_ready(**_: dict) -> None:
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "meshdb.settings")
 
 # Use the docker-hosted Redis container as the backend for Celery
-app = Celery("meshdb", broker=os.environ.get("CELERY_BROKER", "redis://localhost:6379/0"))
+app = Celery("meshdb", broker= environment.CELERY_BROKER)
```
WillNilges, line 68: You should just be able to refer to this by the environment name, since you're importing it directly.

```suggestion
app = Celery("meshdb", broker=CELERY_BROKER)
```

PR #873, opened 2025-03-27
Add InstallFeeBillingDatum model

### LINKED ISSUE #828: Add billing component
Our building contracts limit billing to once per unit, but over time new tenants move in and request new installs, therefore mesh billing records needs to associate with the unit and not the tenant.

So, in db terms, I think the `billing status` mentioned below needs to exist outside of the Install entity.

To achieve this, I suggest a simple table adjustment and table addition.

Install entity
• add `billable` column. boolean value.

new Billing Table columns…
• `building#`
• `unit#`
• `billing status`

When `billable` is first set to “TRUE”, Billing Table row is created (if it doesn't already exist) and `billing status` should default to “to be billed”. I would then change that value to “billed” after invoicing is complete.

Where Install entity can have multiple rows with the same building and unit numbers (multiple tenants), Billing Table will have only one row with these values (the unit).

Andrew-Dickinson:
Adds a new model, `InstallFeeBillingDatum` to track information related to institutional billing of install fees. Includes model history, API access, api docs, admin page, and an inline on the install admin page. 

Also adds the capability to the nested reference serializer to be permissions aware so that we don't expose the billing data to anyone with install read permissions (just a reference to it via UUID)

I added a separate group to manage the permissions of this model, since Jason requested that editing be more restricted in an attempt to preserve data quality. Open to discussion on this and if it is appropriately principled, but of course it is easier to add more permissions later, than to take them away. This is also the justification for making a separate model rather than just adding additional columns to `Install`, it lets us control the permissions more granularly.

Install inline:
Screenshot 2025-03-27 at 23 26 01

Edit page:
Screenshot 2025-03-27 at 23 27 24

Closes #828


src/meshapi/models/billing.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    billing_date = models.DateField(
+        default=None,
+        blank=True,
+        null=True,
+        help_text="The date that the associated install was billed to the responsible organization",
+    )
+
+    invoice_number = models.CharField(
```
WillNilges, line 49: Is this some number specific to Stripe?

src/meshapi/urls.py
```diff
@@ -30,6 +30,12 @@
     path("sectors/<uuid:pk>/", views.SectorDetail.as_view(), name="meshapi-v1-sector-detail"),
     path("accesspoints/", views.AccessPointList.as_view(), name="meshapi-v1-accesspoint-list"),
     path("accesspoints/<uuid:pk>/", views.AccessPointDetail.as_view(), name="meshapi-v1-accesspoint-detail"),
+    path("billing/installfeedata/", views.InstallFeeBillingDataList.as_view(), name="meshapi-v1-installfeedata-list"),
+    path(
+        "billing/installfeedata/<uuid:pk>/",
```
WillNilges, line 35: should we slug this?
```suggestion
        "billing/install-fee-data/<uuid:pk>/",
```

src/meshapi/models/billing.py
Andrew-Dickinson: I believe it is actually Intuit Quickbooks

src/meshapi/admin/models/install.py
```diff
@@ -181,5 +182,17 @@ def get_node_status(self, obj: Install) -> str:
...
+        static_inlines = super().get_inline_instances(request, obj)
+
+        if (
+            obj
+            and hasattr(obj, "install_fee_billing_datum")
+            and request.user.has_perm("meshapi.view_installfeebillingdatum", None)
+        ):
+            return static_inlines + [InstallFeeBillingDatumInline(self.model, self.admin_site)]
+
+        return static_inlines  # Hide billing inline if no related objects exist
```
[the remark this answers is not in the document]
Andrew-Dickinson: You go via the Add flow for the billing datum model. It’s not perfect but I think it’s very reasonable. Jason is probably going to end up using CSVs to do bulk imports most of the time anyway

src/meshapi/urls.py
Andrew-Dickinson: Is this our first multi-word path? I don’t care either way as long as we’re consistent

src/meshapi/admin/models/install.py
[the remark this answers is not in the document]
WillNilges: ok cool, make sure it's documented in the wiki

src/meshapi/models/billing.py
WillNilges: oh dear god

src/meshapi/urls.py
WillNilges: i don't feel terribly strongly about it

WillNilges (approved):

src/meshapi/admin/models/install.py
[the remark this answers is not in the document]
Andrew-Dickinson: https://wiki.nycmesh.net/books/6-services-software/page/recording-the-invoice-of-an-organization-for-install-fees

src/meshapi/urls.py
Andrew-Dickinson: We're doing kebab case in other places. Good catch

PR #883, opened 2025-04-23
Add Stripe Subscription ID to Install object

Andrew-Dickinson:
Adds a stripe subscription ID to the install object, along with an "eyeball" link in the admin UI when it is present


src/meshapi/admin/models/install.py
```diff
@@ -47,6 +48,10 @@ class Meta:
                 lambda ticket_number: f"{OSTICKET_URL}/scp/tickets.php?number={ticket_number}",
                 title="View in OSTicket",
             ),
+            "stripe_subscription_id": ExternalHyperlinkWidget(
+                lambda subscription_id: STRIPE_SUBSCRIPTIONS_URL + subscription_id,
```
WillNilges, line 52: Consider using an f string

WillNilges (approved):
Consider testing an invalid subscription id

src/meshapi/admin/models/install.py
Andrew-Dickinson: no

PR #892, opened 2025-05-06
Add a post_save hook to keep Stripe metadata synced to MeshDB

Andrew-Dickinson:
Adds a `post_save` hook so that when our pointer to the stripe subscription ID is changed, we tell stripe about it. This makes it much easier for humans to do the reverse mapping of "what install number does this subscription correspond to"

In the event of a conflict, MeshDB is considered the source of truth and we quietly update Stripe to maintain consistency. However this should be super rare, because editing this metadata is not possible via the stripe UI, and we are the only tool editing it

Also rearranged the `.env` file to hopefully make it a bit less confusing for newcomers

Also edits the reconciliation script to add a basic regex parser for JSON exports from the `#donations` slack channel, since this appears to be the best source of truth we have for mapping subscription to install number

TODO:

- [ ] Do integration testing with the live Stripe API after I get keys from Olivier
- [ ] Prep data for MeshDB backfill so that after this merges we can populate all the existing subscriptions


scripts/reconcile_stripe_subscriptions.py
```diff
@@ -27,32 +33,133 @@ def find_install_numbers_by_stripe_email(stripe_email: str) -> List[str]:
...
+                        if install_number_pound_match:
+                            slack_charge_mapping[stripe_charge_id] = install_number_pound_match.group(1)
+                            continue
+
+                        install_number_match = re.search(
+                            r"(?<!NY\s)(?<!NY,\s)(?<!Address:\s)(?<!Address:\s\s)\b(\d{4,5})\b", replies_joined
+                        )
+                        if install_number_match:
+                            install_number = install_number_match.group(1)
+                            if install_number not in [
```
WillNilges, line 105: Can you add a comment or something to explain these magic numbers?

src/meshapi/util/events/update_stripe_subscription.py
```diff
@@ -0,0 +1,156 @@
...
+        time.sleep(1)
+
+    raise RuntimeError(
+        f"Unable to add install {install_number} to Stripe subscription_id {subscription_id}"
+    ) from stripe_error
+
+
+@receiver(post_save, sender=Install, dispatch_uid="update_stripe_subscription")
+@skip_if_flag_disabled("INTEGRATION_ENABLED_UPDATE_STRIPE_SUBSCRIPTIONS")
+@advisory_lock("update_stripe_subscription")  # TODO: Test this race condition
```
WillNilges, line 116: Take care of this TODO?

WillNilges (approved):

PR #898, opened 2025-05-20
Don't schedule celery on the control plane!

WillNilges:
This adds an affinity that should prevent celery pods from getting deployed on the control plane. If this works, I'll do it with our other "heavy" pods too.


james-otten (approved):

james-otten:
Perhaps also the db, feel free to split that out

PR #913, opened 2025-06-08
Add ability to search for additional members in query form

### LINKED ISSUE #912: Query and MeshDB search/show additionnal member on an Install
We have now the option to add a new member (spouse, roommate, etc..) to an Install.
Unfortunately that member(s) does not show when we do a query on the addition member name or its email address.
If it happens you know the install number or NN, the query those not list that additional member.
When we do a member search in MedhDB it founds it but does not show to which install it is linked to.

Pretty useless when receiving a request for support and we try to figure our where that member is located, to what install it is linked to.

The goal of adding the additional member is to figure out which install, where is that member. 

Note: if you do a search of the install number (provided you know it) it shows the additional member.  So it is in there.

Andrew-Dickinson:
Adds the ability to search for fields on the additional members attached to an install. However, in order to make the response make sense, we need a way to expose the additional members in the response.

My solution to this was to replicate the appearance of a SQL join, but with Django model objects. If your query hits an install that has additional members, it will show up in the output multiple times. So it will look something like

```
#1234, John Smith, NN123....
#1234, Jane Doe, NN123....
```

And this "block" of two rows will be returned for queries to `NN123`, `#1234`, or either `John Smith` or `Jane Doe` it effectively moves the unit of query from the install to a "fake group" of installs that are duplicated based on the number of additional members.

**Is this a good idea?** 
_Query form, bottom text_

Closes #912

### DIFF
--- a/src/meshapi/tests/test_query_form.py
+++ b/src/meshapi/tests/test_query_form.py
@@ -21,12 +21,17 @@ class TestQueryForm(TestCase):
         member.save()
         sample_install_copy["member"] = member
 
+        self.additional_member = Member(**sample_member)
+        self.additional_member.name = "Jane Doe"
+        self.additional_member.save()
+
         node = Node(latitude=0, longitude=0, status=Node.NodeStatus.ACTIVE)
         node.save()
 
         self.install = Install(**sample_install_copy)
         self.install.node = node
         self.install.save()
+        self.install.additional_members.add(self.additional_member)
 
     def query(self, route, field, data):
         code = 200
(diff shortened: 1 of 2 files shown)


WillNilges (changes requested):
I don't think this is a good idea. The query form is insecure and deprecated. Can we formulate a better query with the Query Explorer?

OlivierMorf:
This is becoming more urgent. People are requesting support (posting in support) that we don't know who they are/where.

OlivierMorf:
I added that member "Marius Courcoul" as they are subletting the unit, but the install is under Hunderson Sabbat
#9377 NN219

Andrew-Dickinson:
Tracking that bug here: https://github.com/nycmeshnet/meshdb/issues/980

Issue #639, opened 2024-10-10
Add node-explorer data to SQL explorer


Andrew-Dickinson:
Make node explorer data available in sql-explorer so that for example, you can join dependent nodes in with queries of meshdb metadata.

I think we can use [psql-http](https://github.com/pramsey/pgsql-http) to create a stored procedure or `VIEW` which queries node-explorer / outage-analyzer, and provides the result in tabular format

Andrew-Dickinson:
https://nycmesh.slack.com/archives/C06GW3Q92TY/p1728523132885009?thread_ts=1728520828.456049&cid=C06GW3Q92TY

Andrew-Dickinson:
Could also do some kind of ETL job as a celery worker that fills up a table with the data

PR #955, opened 2025-06-29
Add pgsql-http extension to our database

WillNilges:
Will need to `CREATE EXTENSION http` when this is merged.

Candidate for closing https://github.com/nycmeshnet/meshdb/issues/639

This could also increase our attack surface quite a bit. I wonder if there is a way to restrict what URLs we can query.


WillNilges:
https://github.com/pramsey/pgsql-http/issues/193

PR #964, opened 2025-07-01
add links to download KML files

WillNilges:
Adds two new links to the homepage that lets you download a KML that is updated "live" by meshdb, as well as one that links to your local copy of meshdb. The existing KML was renamed to "KML Download (Offline)"

image

### DIFF
--- a/src/meshweb/views/index.py
+++ b/src/meshweb/views/index.py
@@ -29,7 +29,10 @@ def index(request: HttpRequest) -> HttpResponse:
             ("/api/v1/", "MeshDB Data API"),
             ("/api-docs/swagger/", "API Docs (Swagger)"),
             ("/api-docs/redoc/", "API Docs (Redoc)"),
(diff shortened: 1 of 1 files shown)


src/meshweb/views/index.py
```diff
@@ -16,7 +16,8 @@ def index(request: HttpRequest) -> HttpResponse:
         ],
         ("meshweb/volunteer.png", "Volunteer Tools"): [
             ("/admin", "Admin Panel"),
-            ("/api/v1/geography/whole-mesh.kml", "KML Download"),
+            ("https://raw.githubusercontent.com/nycmeshnet/meshdb/refs/heads/main/sampledata/meshdb.kml", "KML Download"),
```
james-otten, line 19: Stale data under with the old link name is confusing.

Andrew-Dickinson: That’s not what this is. Check the contents of that file, it’s a 1kb hyperlink pointer to the latest data

james-otten: Wtf

james-otten: No need to reference github https://github.com/nycmeshnet/meshdb/pull/969

PR #992, opened 2025-08-01
Notify admins when member submitting join form rejects our changes

WillNilges:
As a "Part 1" to our join form logic changes, let's start notifying admins when folks reject our changes. Right now, we still accept our changes even if they use "trust me bro". This indicates that our system did something they percieve as wrong, so a human needs to look at it and maybe reach out to the member to figure it out.


src/meshapi/views/forms.py
```diff
@@ -413,6 +413,18 @@ def process_join_form(r: JoinFormRequest, request: Optional[Request] = None) ->
 
     if r.trust_me_bro:
         logging.warning(success_message)
+        if changed_info:
+            notify_string = "A new member rejected our changes to their address.\n"
+            "Please review the submission and ensure building information is accurate.\n"
+            "You might need to reach out to them and confirm their details\n"
```
Andrew-Dickinson, line 419: Can we emphasize that "due to a MeshDB bug" it is _extremely_ likely that human action is required in order for the database to be correct?

Andrew-Dickinson (approved):

PR #1000, opened 2025-08-08
NYC Address Parsing refactor

WillNilges:
This PR
- Adds retries from the requests library to reduce indirection higher up
- Removes a couple of unnecessary (confusing) exceptions
- Improves(?) variable names
- Adds some comments
- Moves logic out of the NYC address constructor for clarity
- Adds a new type of exception, replacing ValueError with UnsupportedAddress error for clarity (working within the constraints of the current system)

The philosophy I've landed on is: Log it when it happens, be able to tell the user if something unexpected happened (an API is down), or if their input was invalid. Beyond that, we can look at the logs to learn more.

I don't think telling users exactly why their request failed is useful to them. (Do they need to know that our internal zip code handling is a mess?) We should ask them to contact support. Then again, I don't have data on how many people were able to fix their own issues based off what we told them.

Having the actual errors logged separately from our known errors will also make our logs more concise, making it easier to debug known errors, and make errors we haven't seen before or anticipated stick out way more (since they'll still raise exceptions we see).

For a future PR (this one is already too big):
- [ ] Further reduce usage of exceptions
- [ ] Add traces
- [ ] Move NYC address parsing logic out of the constructor entirely? This feels like an antipattern to me that I regret the more I look at it.

Best reviewed commit-by-commit.

### DIFF
--- a/src/meshapi/validation.py
+++ b/src/meshapi/validation.py
@@ -103,13 +103,15 @@ class NYCAddressInfo:
         logging.info(f"Constructing NYCAddressInfo for '{street_address}, {city}, {state} {zip_code}'...")
         self.session = get_requests_session_with_retries()
         if state != "New York" and state != "NY":
-            logging.error(f"State '{state}' is not New York.")
-            raise UnsupportedAddressError
+            error = f"State '{state}' is not New York."
+            logging.error(error)
+            raise UnsupportedAddressError(error)
 
         # This method is only compatible with the 5 boroughs of NYC
         if not NYCZipCodes.match_zip(zip_code):
-            logging.error(f"Zip code '{zip_code}' does not appear in our zip code database.")
-            raise UnsupportedAddressError
+            error = f"Zip code '{zip_code}' does not appear in our zip code database."
+            logging.error(error)
+            raise UnsupportedAddressError(error)
 
         self.address = f"{street_address}, {city}, {state} {zip_code}"
 
@@ -127,12 +129,14 @@ class NYCAddressInfo:
             )
             nyc_geosearch_resp = json.loads(nyc_planning_req.content.decode("utf-8"))
         except Exception:
-            logging.exception("An exception occurred while querying geosearch.planninglabs.nyc")
-            raise AddressAPIError
+            error = "An exception occurred while querying geosearch.planninglabs.nyc"
+            logging.exception(error)
+            raise AddressAPIError(error)
 
         if len(nyc_geosearch_resp["features"]) == 0:
-            logging.error("Address not found in geosearch.planninglabs.nyc.")
-            raise InvalidAddressError
+            error = "Address not found in geosearch.planninglabs.nyc."
+            logging.error(error)
+            raise InvalidAddressError(error)
 
         addr_props = nyc_geosearch_resp["features"][0]["properties"]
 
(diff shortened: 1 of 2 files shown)


src/meshapi/validation.py
```diff
@@ -184,54 +198,81 @@ def __init__(self, street_address: str, city: str, state: str, zip_code: str):
...
                 params=query_params,
                 timeout=DEFAULT_EXTERNAL_API_TIMEOUT_SECONDS,
             )
-            nyc_dataset_resp = json.loads(nyc_dataset_req.content.decode("utf-8"))
+            nyc_dataset_req.raise_for_status()
+        except Exception:
+            logging.exception("[BUILDING_FOOTPRINTS] Exception raised during HTTP Request.")
+            return INVALID_ALTITUDE
+
+        nyc_dataset_resp = json.loads(nyc_dataset_req.content.decode("utf-8"))
```
WillNilges, line 224: move this into the exception?

WillNilges:
I also decided that I hate indirection so I am trying to minimize it.

dgonzdev:
just wondering if the validation pattern here is a Django pattern we're following?

dgonzdev:
random refactoring ideas for the backlog (aka not to do now but later):
- maybe some of these init methods with business logic that create objects could be refactored to factory class methods? 
- maybe the api calls could be pulled into separate clients? even seems like there's a chance to create an open source library for pinging City of NYC apis.

WillNilges:
> just wondering if the validation pattern here is a Django pattern we're following?

Good question. This is not explicitly a Django pattern we're following, no. I'll be honest: I don't know what I was thinking when I set this up. I was trying to be "clever" with the use of Exceptions but I think it blew up in my face.

> random refactoring ideas for the backlog (aka not to do now but later):
> 
>     * maybe some of these init methods with business logic that create objects could be refactored to factory class methods?

That's what I was thinking too.
 
>     * maybe the api calls could be pulled into separate client classes? even seems like there's a chance to create an open source library for pinging City of NYC apis.

I hadn't thought of that. I like the idea of classes, but writing a whole library is a little outside my area of interest.

src/meshapi/validation.py
```diff
@@ -95,6 +97,16 @@ class NYCAddressInfo:
     bin: int | None
 
     def __init__(self, street_address: str, city: str, state: str, zip_code: str):
+        # We're going to be making a lot of HTTP requests. It would be better
+        # to retry them
+        retries = Retry(total=3,  # Total number of retries
+                        backoff_factor=1,  # Wait time between retries
+                        status_forcelist=[500, 502, 503, 504])  # HTTP status codes to retry
```
Andrew-Dickinson: Any way to do >500 here?

Andrew-Dickinson: Or perhaps we want to retry anything 4xx or 5xx

src/meshapi/util/requests.py
```diff
@@ -0,0 +1,15 @@
+from requests import Session
+from requests.adapters import HTTPAdapter, Retry
+
+
+def get_requests_session_with_retries() -> Session:
+    retries = Retry(
+        total=3,
+        backoff_factor=1,
+        status_forcelist=[500, 502, 503, 504],
```
Andrew-Dickinson, line 9: Perhaps include any 4xx or 5xx?

src/meshapi/validation.py
```diff
@@ -95,16 +100,16 @@ class NYCAddressInfo:
     bin: int | None
 
     def __init__(self, street_address: str, city: str, state: str, zip_code: str):
-        full_address = f"{street_address}, {city}, {state} {zip_code}"
+        logging.info(f"Constructing NYCAddressInfo for '{street_address}, {city}, {state} {zip_code}'...")
+        self.session = get_requests_session_with_retries()
         if state != "New York" and state != "NY":
-            raise ValueError(f"(NYC) Address '{full_address}' is unsupported: State '{state}' is not New York.")
+            logging.error(f"State '{state}' is not New York.")
+            raise UnsupportedAddressError
```
Andrew-Dickinson, line 107: Message?

Issue #879, opened 2025-04-06
Add "Node Placement" field on the Node Model


Andrew-Dickinson:
Add a field on the node model which allows us to track the placement of the node within the building it is located on. This lets us filter out balcony installs in some queries, such as when automatically determining if a given join request can skip the panorama step of the intake proces

Possible values:
- Rooftop
- Balcony
- Window
- Indoor (Fiber Sites, non-rooftop Supernodes, etc)
- Unknown (Stored as null in the DB) (Default)

Slack ref: https://nycmesh.slack.com/archives/C06GW3Q92TY/p1743778466388549

Andrew-Dickinson:
@dgonzdev is going to try to take this over

PR #1030, opened 2025-09-08
add placement field to node model

dgonzdev:
https://github.com/nycmeshnet/meshdb/issues/879#issuecomment-3152820825

Add "Node Placement" field on the Node Model

placement


Andrew-Dickinson (approved):
LGTM. If you want bonus points you could update this conditional to only send the NN to OSTicket when the placement is "rooftop" (or unknown I guess)
https://github.com/nycmeshnet/meshdb/blob/main/src/meshapi/util/events/osticket_creation.py#L82

which was the original motivation for this field

Andrew-Dickinson:
Discussed in OH, probably we want a unit test also

Issue #775, opened 2024-12-07
Add link types for WDS and Wlan4


Andrew-Dickinson:
Add link types for WDS and wlan4. Think maybe they should be called: 

- 5 GHz WDS
- 5 GHz Wlan4

Then perhaps we should write a data migration to convert the existing 5 GHz enum to “5 GHz Unspecified” (and add “5GHz Airmax”)

Andrew-Dickinson:
Probably should update the UISP import to correctly import airmax links appropriately

Andrew-Dickinson:
https://nycmesh.slack.com/archives/C06GW3Q92TY/p1733590799683629

Andrew-Dickinson:
https://nycmesh.slack.com/archives/C06GW3Q92TY/p1736261641294889?thread_ts=1735898007.108349&cid=C06GW3Q92TY

WillNilges:
@Tanner on slack will take this on. I'll work with him on this :)

Andrew-Dickinson:
Talked to Loyd F in office hours today and he might attempt to work on this

Andrew-Dickinson:
@dgonzdev is going to look at this one

PR #1035, opened 2025-09-19
add more link types

dgonzdev:
https://github.com/nycmeshnet/meshdb/issues/775


dgonzdev:
@Andrew-Dickinson i had two questions about ticket https://github.com/nycmeshnet/meshdb/issues/775

are the changes in this pr what we wanted? aka are the link types in the pr correct?

and how do we create/run data migrations in mesh db?

PR #1046, opened 2025-10-12
Add more link types attempt 2

dgonzdev:
Add link types for WDS and Wlan4 https://github.com/nycmeshnet/meshdb/issues/775

replaces the first attempt in this pr https://github.com/nycmeshnet/meshdb/pull/1035

### DIFF
--- a/src/meshapi/tests/test_uisp_import.py
+++ b/src/meshapi/tests/test_uisp_import.py
@@ -71,12 +71,12 @@ class TestUISPImportUtils(TestCase):
 
         self.assertEqual(
             get_link_type({"type": "wireless", "frequency": 5000}),
-            Link.LinkType.FIVE_GHZ_UNSPECIFIED,
+            Link.LinkType.FIVE_GHZ_AIRMAX,
         )
 
         self.assertEqual(
             get_link_type({"type": "wireless", "frequency": 5900}),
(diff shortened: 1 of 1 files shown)


src/meshapi/migrations/0014_alter_historicallink_type_alter_link_type.py
```diff
@@ -0,0 +1,59 @@
...
+                null=True,
+            ),
+        ),
+        migrations.AlterField(
+            model_name="link",
+            name="type",
+            field=models.CharField(
+                blank=True,
+                choices=[
+                    ("5 GHz", "5 GHz"),
```
Andrew-Dickinson:
```suggestion
                    ("5 GHz", "5 GHz Unspecified"),
```

Andrew-Dickinson (commented):
Looks good, except that I would also make it clear in the UI that the unspecified type is unspecified

Andrew-Dickinson (commented):
One nit (Willard cares about this more than I do) you can squash the migrations with:
```
 python manage.py squashmigrations meshdb 0014 0015
 ```

Andrew-Dickinson:
Weird, we have everything pinned, which is supposed to prevent build non-determinism like this. Yeah let's go ahead and add ` types-python-dateutil == 2.9.*` to the `dev` section of `pyproject.toml` and see if that fixes it

Issue #1112, opened 2026-03-25
Remove Logic which creates/updates "Existing Link" LOS Objects


Andrew-Dickinson:
The LOS object's history table grows very large in size due to constant date mutations on each daily import, and dramatically increases the size of database backups. This task is to remove the calls that create and update these objects:
https://github.com/nycmeshnet/meshdb/blob/main/src/meshapi/util/uisp_import/sync_handlers.py#L350

They're not used by anyone. Only the manual annotations are used and those are obviously input manually.

Cleaning the old entries out of the DB is out of scope for this issue, we will handle that separately

PR #1118, opened 2026-04-22
remove sync_link_table_into_los_objects calls

dgonzdev:
https://github.com/nycmeshnet/meshdb/issues/1112

```
The LOS object's history table grows very large in size due to constant date mutations on each daily import, and dramatically increases the size of database backups. This task is to remove the calls that create and update these objects:
https://github.com/nycmeshnet/meshdb/blob/main/src/meshapi/util/uisp_import/sync_handlers.py#L350
```


dgonzdev:
@Andrew-Dickinson are the changes in this pr the ask for this ticket or were you thinking something different?

cc @WillNilges

WillNilges (approved):

PR #1120, opened 2026-06-17
Replace py3-validate-email with email-validator

rubendax:
## Why

The [py3-validate-email](https://git.ksol.io/karolyi/py3-validate-email) library has been sabotaged by its author who intentionally inserted a racial slur into it's license. This in turn triggered PyPi, Github, and others, to ban the repo. Needless to say, MeshDB should not continue to use it as a dependency. 

This PR replaces `py3-validate-email` with [`email-validator`](https://github.com/JoshData/python-email-validator) (v2.x), a  library that is used internally by pydantic, fastAPI, and other major Python packages.

### `pyproject.toml`
- Swap `py3-validate-email @ git+https://git.ksol.io/karolyi/py3-validate-email@v1.0.12` → `email-validator==2.*`.

### `src/meshapi/validation.py`
- Replace `validate_email_or_fail()` with `email_validator.validate_email(check_deliverability=True, timeout=5)`.
- Collapse five exception types (`DNSTimeoutError`, `SMTPCommunicationError`, `TLSNegotiationError`, `SMTPTemporaryError`, `EmailValidationError`) into a single `EmailNotValidError` catch.
- Simplify return type from `Optional[bool]` → `bool`.

### `src/meshapi/views/forms.py`
- Remove `try/except EmailValidationError` wrapper around the validation call — no longer needed since `validate_email_address()` now always returns `True`/`False` without raising.
- Remove `from validate_email.exceptions import EmailValidationError`.

### `src/meshapi/tests/test_join_form.py`
- Remove three tests that exercised old exception paths (`DNSTimeoutError` → 500, `SMTPTemporaryError` → 201, `SMTPTemporaryError` + bad email → 201).
- Add one test verifying the simplified contract: `validate_email_address()` returns `False` → view returns 400.

#### Valid email:
before: 201 
after: 201

#### Invalid syntax / undeliverable domain:
before: 400 
after: 400

#### DNS timeout:
before: 500 (re-raised as internal error) 
after: 201 (allowed through)

#### SMTP temporary error:
before: 201 (benefit of the doubt)
after: N/A — no SMTP check performed

The only user-facing behavior change is **DNS timeouts**: previously these returned a 500 error to the user, blocking potentially legitimate submissions. Now the email-validator library treats DNS timeouts as "unknown deliverability" (returns successfully without raising), so the submission proceeds. I feel like this mostly aligns with the pre-existing "benefit of the doubt" philosophy already applied to SMTP issues.

Note: The new library does **not** perform SMTP checks (connecting to the mail server to verify the recipient). That being said, this actually matches our previous library configuration: `check_smtp=False`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ dependencies = [
     "python-dotenv==1.0.*",
     "stringcase==1.2.*",
     "python-dotenv==1.0.*",
-    "py3-validate-email @ git+https://git.ksol.io/karolyi/py3-validate-email@v1.0.12",
+    "email-validator==2.*",
     "phonenumbers==8.13.*",
     "requests==2.32.*",
     "requests-mock==1.12.*",
--- a/src/meshapi/views/forms.py
+++ b/src/meshapi/views/forms.py
@@ -17,7 +17,6 @@ from rest_framework.decorators import api_view, permission_classes
 from rest_framework.request import Request
 from rest_framework.response import Response
 from rest_framework_dataclasses.serializers import DataclassSerializer
-from validate_email.exceptions import EmailValidationError
 
 from meshapi.exceptions import InvalidAddressError, UnsupportedAddressError
 from meshapi.models import AddressTruthSource, Building, Install, Member, Node
@@ -160,18 +159,8 @@ def process_join_form(r: JoinFormRequest, request: Optional[Request] = None) -> 
     if not r.email_address:
         return Response({"detail": "Must provide an email"}, status=status.HTTP_400_BAD_REQUEST)
 
-    try:
-        if r.email_address and not validate_email_address(r.email_address):
-            return Response({"detail": f"{r.email_address} is not a valid email"}, status=status.HTTP_400_BAD_REQUEST)
-    except EmailValidationError:
-        # DNSTimeoutError, SMTPCommunicationError, and TLSNegotiationError are all subclasses of EmailValidationError.
-        # Any other EmailValidationError will be caught inside validate_email_address() and trigger a return false,
-        # so we know that if validate_email_address() throws, EmailValidationError, it must be one of these
-        # and therefore our fault
-        return Response(
-            {"detail": "Could not validate email address due to an internal error"},
-            status=status.HTTP_500_INTERNAL_SERVER_ERROR,
-        )
+    if r.email_address and not validate_email_address(r.email_address):
+        return Response({"detail": f"{r.email_address} is not a valid email"}, status=status.HTTP_400_BAD_REQUEST)
 
     # Expects country code!!!!
     if r.phone_number and not validate_phone_number(r.phone_number):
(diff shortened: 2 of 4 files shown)


WillNilges:
> The only user-facing behavior change is DNS timeouts: previously these returned a 500 error to the user, blocking potentially legitimate submissions. Now the email-validator library treats DNS timeouts as "unknown deliverability" (returns successfully without raising), so the submission proceeds. I feel like this mostly aligns with the pre-existing "benefit of the doubt" philosophy already applied to SMTP issues.

This is probably fine. If the library says it _looks_ valid, that's probably good enough. I vaguely remember seeing DNS timeouts causing join records to fail transiently, but would always pass when I retried them. We should watch out though: If we start seeing invalid emails due to this, we should go back to the old behavior.

src/meshapi/views/forms.py
```diff
@@ -160,18 +159,8 @@ def process_join_form(r: JoinFormRequest, request: Optional[Request] = None) ->
...
-            return Response({"detail": f"{r.email_address} is not a valid email"}, status=status.HTTP_400_BAD_REQUEST)
-    except EmailValidationError:
-        # DNSTimeoutError, SMTPCommunicationError, and TLSNegotiationError are all subclasses of EmailValidationError.
-        # Any other EmailValidationError will be caught inside validate_email_address() and trigger a return false,
-        # so we know that if validate_email_address() throws, EmailValidationError, it must be one of these
-        # and therefore our fault
-        return Response(
-            {"detail": "Could not validate email address due to an internal error"},
-            status=status.HTTP_500_INTERNAL_SERVER_ERROR,
-        )
```
WillNilges, line 174: I'm on the fence about whether we should remove this. I don't want to turn someone away on the join form thinking it was their fault when we failed to correctly check their email.

To be clear: The difference here is that if we _do_ have a problem, we return a 201 anyway and they continue with their submission? If that's the case, that's probably fine.

pyproject.toml
```diff
@@ -15,7 +15,7 @@ dependencies = [
     "python-dotenv==1.0.*",
     "stringcase==1.2.*",
     "python-dotenv==1.0.*",
-    "py3-validate-email @ git+https://git.ksol.io/karolyi/py3-validate-email@v1.0.12",
```
WillNilges, line 18: It's a shame about this library. I hope the author gets the help he needs, but that kind of behavior is not acceptable in our community.

WillNilges (approved):

src/meshapi/views/forms.py
rubendax: The difference is that DNS timeouts would now silently pass instead of raising a 500. email-validator doesn't throw an error on DNS timeouts; it returns an "unknown deliverability" result and the function returns True. So the submission continues with a 201. Basically the same logic that we already had for detecting SMTP issues, just extended to DNS timeouts too. Feels like it would be better to let someone through than to 500 them because of a slow nameserver?

PR #1122, opened 2026-07-14
scramble script improvements + misc fixes

WillNilges:
- Added `--skip-*` flags to selectively skip scrambling individual entity types. In a follow up PR I may change the interface for this because these flags are cumbersome.
- Added more stuff to delete in the scramble script. This will let us use the newer pgdump method of exporting the database.
- Added a fix for the Jefferson Situation, where a node can exist when an install is using its NN as an install #
- Removed debug `print` statement from `src/meshapi/admin/inlines.py:208` that was leaking type info to the console during admin operations
- Added NixOS dev setup instructions to `README.md`, directing NixOS users to activate the environment via `nix-shell`
- Updated `shell.nix` to use Python 3.12 packages instead of Python 3.11 so it actually works. We're still on Python 3.11. need to fix that someday.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ If you would like to develop in a [Dev Container](https://code.visualstudio.com/
 
 ### with Host
 
+If you are using NixOS, activate the dev environment with `nix-shell`, and then proceed with the below instructions.
+
 If you would like to develop directly on your host machine
 
 Install Python 3.11 and pip.
--- a/shell.nix
+++ b/shell.nix
@@ -2,9 +2,9 @@
 (pkgs.buildFHSEnv {
   name = "pipzone";
   targetPkgs = pkgs: (with pkgs; [
-    python311
-    python311Packages.pip
-    python311Packages.virtualenv
+    python312
+    python312Packages.pip
+    python312Packages.virtualenv
   ]);
   runScript = "zsh";
 }).env
--- a/shell.nix
+++ b/shell.nix
@@ -2,7 +2,7 @@
 (pkgs.buildFHSEnv {
   name = "pipzone";
   targetPkgs = pkgs: (with pkgs; [
-    python312
+    python311
     python312Packages.pip
     python312Packages.virtualenv
   ]);
--- a/src/meshapi/admin/inlines.py
+++ b/src/meshapi/admin/inlines.py
@@ -205,7 +205,6 @@ class AdditionalMembersInline(TabularInline):
     show_change_link = True
 
     def name(self, instance: Any) -> str:
-        print(type(instance))
         return instance.member.name if instance.member else "-"
 
     def primary_email_address(self, instance: Any) -> str:
--- a/src/meshapi/management/commands/scramble_members.py
+++ b/src/meshapi/management/commands/scramble_members.py
@@ -9,7 +9,10 @@ from django.core.management.base import BaseCommand
 from django.db import transaction
 from faker import Faker
 
+from django.contrib.auth.models import User
+
 from meshapi.models import LOS, Install, Member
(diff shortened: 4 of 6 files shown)


src/meshapi/models/node.py
```diff
@@ -120,6 +120,11 @@ class NodePlacement(models.TextChoices):
         "should not be relied on by automated systems. ",
     )
 
+    # XXX (wdn):  DO NOT USE in production! This is lubrication for the scramble
```
bradyjoh: hope this is safe to merge

bradyjoh (approved):

src/meshapi/models/node.py
jedahan: can this function on the model check if it is running in production environment and refuse?

README.md
```diff
@@ -44,6 +44,8 @@ If you would like to develop in a [Dev Container](https://code.visualstudio.com/
 
 ### with Host
 
+If you are using NixOS, activate the dev environment with `nix-shell`, and then proceed with the below instructions.
+
 If you would like to develop directly on your host machine
 
 Install Python 3.11 and pip.
```
jedahan, line 51: should this be updated to 3.12?

src/meshapi/models/node.py
WillNilges: I did this to fix this error.

```[2026-07-14 20:15:03,141] [522897] [ERROR] Error saving scrambled Node the normal way. Forcing it...
Traceback (most recent call last):
  File "/home/wilnil/Code/nycmeshnet/meshdb/src/meshapi/management/commands/scramble_members.py", line 173, in handle
    node.save()
  File "/home/wilnil/Code/nycmeshnet/meshdb/src/meshapi/models/node.py", line 159, in save
    validate_network_number_unused_and_claim_install_if_needed(self.network_number, self.id)
  File "/home/wilnil/Code/nycmeshnet/meshdb/src/meshapi/util/network_number.py", line 62, in validate_network_number_unused_and_claim_install_if_needed
    raise ValueError(
ValueError: Invalid NN: 5933 has an install associated that looks active (#5933)

```

Going to see if I can take another crack at this because the tests are failing anyway.

WillNilges: I think this is happening because the install is active, but the node is not. This might be a valid discrepency in the data. Gonna check in on slack about it.

README.md
WillNilges: Yes, but I am not sure I want to do that in this PR. Looks like NixOS has moved on from Python 3.11, but I should be able to use python311 and then refer to the pip and venv packages in python312Packages.

src/meshapi/models/node.py
WillNilges: OK, I think I found a better way of doing this.

bradyjoh (approved):

shell.nix
```diff
@@ -3,8 +3,8 @@
   name = "pipzone";
   targetPkgs = pkgs: (with pkgs; [
     python311
-    python311Packages.pip
-    python311Packages.virtualenv
+    python312Packages.pip
```
rubendax, line 6: Are we intentionally bumping pip and virtualenv to 3.12 even though we are keeping the interpreter at 3.11?

rubendax: oh, sorry, just saw that this was addressed in another comment. So it works to mix versions like this? I don't know anything about NixOS.

WillNilges: Yeah this is ~fine. It has no impact on production. This is just for developer tooling, and AFAIK I'm the only Nix user developing MeshDB right now, so really this only affects me for now :^)

rubendax (approved):

